PR #18, opened 2019-09-30
Cohort characterization

gowthamrao:
Added functionality to existing functions to make it easier to create reports. 
Feature analyses details adds details about features

### DIFF
--- a/R/CohortDefinition.R
+++ b/R/CohortDefinition.R
@@ -502,10 +502,10 @@ invokeCohortSetGeneration <- function(baseUrl, sourceKeys, definitionIds) {
   df
 }
 
-#' Get cohort inclusion rules and person counts
+#' Get cohort inclusion rule counts for events and persons
 #'
 #' @details
-#' Obtains the inclusion rules from a cohort definition and summarizes the person counts per rule
+#' Obtains the inclusion rules from a cohort definition and summarizes the event and person counts per rule
 #'
 #' @param baseUrl         The base URL for the WebApi instance, for example:
 #'                        "http://server.org:80/WebAPI".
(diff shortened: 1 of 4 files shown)


R/CohortDefinition.R
```diff
@@ -516,19 +516,46 @@ invokeCohortSetGeneration <- function(baseUrl, sourceKeys, definitionIds) {
...
+  })
+  inclusionRuleStatsByPersonDf <- do.call(rbind.data.frame, inclusionRuleStatsByPersonDf)
+  
+  resultsByEvent <- list(summary = summaryByEventDf, 
+                         inclusionRuleStats = inclusionRuleStatsByEventDf
+  )
+  resultsByPerson <- list(summary = summaryByPersonDf, 
+                          inclusionRuleStats = inclusionRuleStatsByPersonDf
+  )
+  results <- list(resultsByEvent = resultsByEvent, resultsByPerson = resultsByPerson
```
alondhe: To follow convention, remove the "results <-" part

R/WebApi.R
```diff
@@ -136,6 +136,8 @@ getCdmSources <- function(baseUrl) {
     }
     data.frame(sourceName = s$sourceName,
                sourceKey = s$sourceKey,
+               sourceNameShort = gsub("\\s*\\([^\\)]+\\)","",as.character(s$sourceName)),
```
alondhe: @gowthamrao -- I don't think these fields necessarily provide value for sites outside of Janssen. For instance, only at Janssen is there a source version (load id) in the name. 

I think we should leave these out, as they are not going to be helpful for other sites, and may instead cause confusion.

R/CohortDefinition.R
gowthamrao:
```suggestion
  list(resultsByEvent = resultsByEvent, resultsByPerson = resultsByPerson
```

schuemie:
I recommend making all changes against the develop branch, not the master branch.

PR #28, opened 2020-01-06
added functionality to extract data sources into R using webapi

gowthamrao:
This PR adds ability to extract pre-computed achilles output into R. Once in R, we can render custom graphs/tables based on research needs. E.g. it will now be able to create a r-markdown report that describes the number of people in the data source with certain condition/procedure etc.


R/WebApi.R
```diff
@@ -142,3 +142,39 @@ getCdmSources <- function(baseUrl) {
...
+#' @examples
+#' \dontrun{
+#' #
+#'
+#' getConfiguration(baseUrl = "http://server.org:80/WebAPI", sourceKeys = c('source1', 'source2'))
+#' }
+#'
+#' @export
+getConfiguration <- function(baseUrl,sourceKeys){
+  require(dplyr)
```
schuemie: Do not use `require` or `library` in a function. Instead, rely on the double-colon notation to call functions in an imported package, e.g.
```r
dplyr::filter()
```
Otherwise your function will have unintended side-effects, in this case loading the dplyr package in the user's search path, which could lead to problems later on.

Also, make sure the dependency is in the import section of the DESCRIPTION file. I think you just deleted dplyr there?

R/WebApi.R
```diff
@@ -142,3 +142,39 @@ getCdmSources <- function(baseUrl) {
...
+#'
+#' @examples
+#' \dontrun{
+#' #
+#'
+#' getConfiguration(baseUrl = "http://server.org:80/WebAPI", sourceKeys = c('source1', 'source2'))
+#' }
+#'
+#' @export
+getConfiguration <- function(baseUrl,sourceKeys){
```
schuemie: This function name is very vague. I literally have no idea what it will return. Your documentation says the function will return 'A list representing the WebApi configuration for the CDM Schema'. Could you be more specific? What would I need this for? Maybe in the example show how the result would be used, for example printing the list of CDM sources (if that is what it returns).

R/WebApi.R
```diff
@@ -142,3 +142,39 @@ getCdmSources <- function(baseUrl) {
...
+  require(dplyr)
+  .checkWebApiBaseUrl(baseUrl)
+  
+  a <- .getCdmSources(baseUrl)
+  a$parsed <- a$parsed %>%
```
schuemie: I'm pretty sure the `%>%` notation requires the dplyr package to be loaded on the search path, which we don't want to do. Use :: notation instead.

R/webApiDataSources.R
```diff
@@ -0,0 +1,118 @@
...
+#' Get the results from individual Achilles reports by data Source definition expression
+#'
+#' @details
+#' Obtain the JSON expression from WebAPI for a given dataSource
+#'
+#' @param baseUrl        The base URL for the WebApi instance, for example:
+#'                       "http://server.org:80/WebAPI".
+#' @param sourceKey      The source key for a CDM instance in WebAPI, as defined in the
+#'                       Configuration page
+#' @param reportType     The precomputed report type in Atlas (Achilles report e.g. dashboard, datadensity, person etc)
```
schuemie: Provide an exhaustive list of the report types, and add code that checks that the user input is one of the allowed options.

R/webApiDataSources.R
```diff
@@ -0,0 +1,118 @@
...
+#' @return
+#' A list of object with pre computed Achilles results for the sourceKey and Achilles report
+#'
+#' @examples
+#' \dontrun{
+#' # This will obtain pre-computed summary regarding dataSource
+#'
+#' getDataSourceReportForSourceKey(baseUrl = "http://server.org:80/WebAPI", dataSource = 'dataSource', reportType = 'dashboard')
```
schuemie: What is the 'data source report'? How would one use the JSON that comes out of this function?

R/webApiDataSources.R
```diff
@@ -0,0 +1,118 @@
...
+#' }
+#'
+#' @export
+
+getDataSourceReportForSourceKey <- function(baseUrl,
```
schuemie: What is the source key, and why is it important enough to mention in the function name?

R/webApiDataSources.R
```diff
@@ -0,0 +1,118 @@
...
+  .checkWebApiBaseUrl(baseUrl)
+
+  #validate source keys
+  sourceKeys <- .getValidSourceKeys(baseUrl, sourceKeys)
+
+  reportTypes <- c('dashboard', 'datadensity', 'person', 'visit', 'condition', 'conditionera',
+                      'procedure', 'drug', 'drugera', 'measurement', 'observation', 'death', 'achillesheel'
+                      )
+
+  writeLines(paste0('Downloading Achilles data into R'))
```
schuemie: paste0 not needed

R/webApiDataSources.R
```diff
@@ -0,0 +1,118 @@
...
+#' @examples
+#' \dontrun{
+#' # This will obtain pre-computed summary regarding dataSource
+#'
+#' getDataSource(baseUrl = "http://server.org:80/WebAPI", cdmDatabaseSchemas = c('schema1','schema2')
+#' }
+#'
+#' @export
+
+getDataSource <- function(baseUrl,sourceKeys){
```
schuemie: Again, very vague function name. Why not `getAchillesData` or similar?

R/webApiDataSources.R
```diff
@@ -0,0 +1,118 @@
...
+
+
+      jsonParsed <- result[[sourceKey]][[reportType]]$parsed
+
+      jsonNames <- names(jsonParsed)
+
+      if (is.null(jsonNames)) {
+        result[[sourceKey]][[reportType]]$parsed <- dplyr::bind_rows(jsonParsed)
+      } else {
+          for (k in (1:length(jsonNames))) {
```
schuemie: Triple-nested for-loop: consider vectorizing

R/WebApi.R
gowthamrao: Thank you! I must have substituted dplyr with tidyverse. This probably is not best practice anyways. I have put dplyr back.

R/webApiDataSources.R
gowthamrao: To Do (will attempt in future?)

R/WebApi.R
schuemie: Just make sure the function works without requiring the user to load magitr on their search path (in other words, without requiring the user to type `library('magitr')` before running your function)

R/webApiDataSources.R
schuemie: I don't understand. How would you 'recreate another Achilles like instance'? What does that even mean?

R/webApiDataSources.R
schuemie: 'getDataSourceReport'? Is there another way to identify a data source (not using the source key)? Why is it important to use the source key anyway? Wouldn't it make more sense to use the other function (getSourceKeyConfiguration) to find the data source currently linked to the source key, and then use this function to get the data source report for that data source?

gowthamrao: The API calls for data-source reports (achilles reports) are dependent on sourceKey (sourceKey is used as the alias). The call is of the form.
`url <- sprintf("%s/cdmresults/%s/%s", baseUrl, sourceKey, reportType)`

So, to be unambiguous in naming and to make the API call we have to provide baseUrl + sourceKey + reportType - we have to include sourceKey.

R/webApiDataSources.R
gowthamrao: changed to paste

R/webApiDataSources.R
gowthamrao: There are many reasons to use the raw Json. 

1. having the raw and full json archived in the study folder, allows us to go back and restore it in another instance of Achilles. Why is this important: lets say two years from now, we were asked why certain design choices were made (or why alternatives were not considered). If the answer to this question was not already documented, then we could go to the restored achilles
2. the parsed data objects derived by parsing JSON into R-friendly objects like data frames, are based on parsing logic. Our parsing logic is not 100% reversible to JSON i.e. we cannot create an Achilles instance from the parsed data.

R/WebApi.R
gowthamrao: I dont know the answer. if dplyr is being used in the package, then dplyr is supposed to re-export magitr and attach it. dplyr is required for the package. 

i will do some research.

gowthamrao: @schuemie I think we can address this by adding magitr to 'Depends:' section of DESCRIPTION file. 
I made the changes, and loaded the package - and found that magitr was attached.

image

R/private.R
```diff
@@ -0,0 +1,187 @@
...
+
+# converts time in integer/milliseconds to date-time with timezone.
+# assumption is that the system timezone = time zone of the local server running webApi.
+.millisecondsToDate <- function(milliseconds) {
+  sec <- milliseconds/1000
+  as.POSIXct(sec, origin = "1970-01-01", tz = Sys.timezone())
+}
+
+# checks if url conforms with expected structure for base url
+.checkWebApiBaseUrl <- function(baseUrl){
```
alondhe: Why the new function? Was the existing one insufficient? If so, let's modify the existing function (.checkBaseUrl).

R/webApiDataSources.R
```diff
@@ -0,0 +1,126 @@
+# @file dataSources
```
alondhe: Rename file to DataSources.R to be consistent with other file names

R/private.R
```diff
@@ -0,0 +1,187 @@
...
+  .getCdmSources(baseUrl) %$%
+    parsed %>%
+    dplyr::filter(toupper(sourceKey) %in% toupper(sourceKeys)) %>%
+    dplyr::select(sourceKey) %>%
+    pull()
+}
+
+
+# get CDM sources
+.getCdmSources <- function(baseUrl) {
```
alondhe: I'd prefer we have 1 getCdmSources function rather than one public and now this private one. Seems like we could just merge the logic into the former.

R/webApiDataSources.R
gowthamrao: thanks. agreed.

R/private.R
```diff
@@ -0,0 +1,187 @@
...
+    result <- list(
+      native = native,
+      parsed = parsed
+    )
+    result
+  }
+}
+
+# get valid source keys in webapi
+.getValidSourceKeys <- function(baseUrl,sourceKeys){
```
alondhe: This function is written twice

R/webApiDataSources.R
alondhe: @schuemie do you mean using lapply?

R/private.R
gowthamrao: sure - lets deprecate the original function and replace with this one. I will rename the new one to .checkBaseUrl. I will comment out the old one - we can delete it later.

R/private.R
gowthamrao: the general thought is that the public function needs to be deprecated and replaced with private function - because this function does not need to be end-user facing.

R/private.R
gowthamrao: resolved

R/private.R
gowthamrao: the two functions are actually different. the public function getCdmSources returns cdmDatabaseSchema,  vocabDatabaseSchema, resultsDatabaseSchema - because these assigned in the function.

```
    cdmDatabaseSchema <- NA
    vocabDatabaseSchema <- NA
    resultsDatabaseSchema <- NA

```

The new private function, gets all the configuration information. It uses a transpose logic to create column names. column names are derived from field 'daimonType' in daimon table. The daimon types are CDM, Vocabulary, Results 

```
  daimons <- result$parsed$daimons %>%
    dplyr::bind_rows() %>%
    tidyr::spread(key = daimonType, value = tableQualifier) %>%
    dplyr::select(-sourceDaimonId) %>%
    base::replace(is.na(.), "") %>%
    dplyr::group_by(sourceId) %>%
    dplyr::summarise_all(max) %>%
    dplyr::mutate_all(na_if,"")
```
I recommend we retain this new function and the old one. We can think of a new name for new function.

gowthamrao: How about getSourceAndDaimonConfiguration ?

R/WebApi.R
schuemie: Yes, I know that would solve it ;-) But I'm explicitly saying that is not an option: we want to minimize the side-effects of running a package and running a function, meaning we should avoid changing the user's search path if possible. If you look at the other OHDSI R packages, the only things that you may find in the Depends section are DatabaseConnector and sometimes Cyclops, because the user must have those on the search path to use the functions in the package. For example, the use must use createConnectionDetails to use CohortMethod. Anything that is not absolutely required to use the package should be in the Imports section, meaning we can use it in the package, but it does not get loaded on the search path. That means we need to use the :: to reference the package functions, which gets a bit tricky for magrittr's weird operators. It is still possible, but you should write `magrittr::`%>%`(a, b)` instead of `a  %>% b`.

R/webApiDataSources.R
schuemie: I'm aware of the need to be reproducible. But I would still like to know how this would be done. Do you mean there's another WebApi function that takes this JSON as input and creates a new Achilles instance? I thought Achilles nowadays is stored as database tables, not as JSON objects?

The reason I'm pushing back on this is because I don't want sham reproducibility, where we store a massive text blob and say we're reproducible, while there's no way to use that blob. 

I'm also not sure whether storing all Achilles data study-side is the solution. We don't do the same thing for CDM data: If I run a study, I store a reference to the exact CDM database I used, and rely on our DB admins to make sure in the future we can retrieve that CDM database using the reference. I don't store a full copy of the CDM database with my study. Why use different behavior here?

gowthamrao:
closing this pull request, and archiving this conversation because it has a lot of changes. will break it into smaller pull request and restart the conversation

R/WebApi.R
gowthamrao: thanks martijn! we will follow this

R/webApiDataSources.R
gowthamrao: makes sense. we could revisit this thought when there is a need to create a new full/partial Achilles (like) instance. For now, i agree with you - lets not call ourselves reproducible by making copies of blobs of text.!

R/WebApi.R
gowthamrao: @schuemie using `magrittr::%>%(a,b)` would make this practically not usable. 

I dont think it is reasonable to translate an easy to follow code like below to the form `magrittr::%>%(a,b)` .
```

    daimons <- result$parsed$daimons
	daimons
      dplyr::bind_rows() %>%
      tidyr::spread(key = daimonType, value = tableQualifier) %>%
      dplyr::select(-sourceDaimonId) %>%
      base::replace(is.na(.), "") %>%
      dplyr::group_by(sourceId) %>%
      dplyr::summarise_all(max) %>%
      dplyr::mutate_all(dplyr::na_if,"") %>% 
      dplyr::rename(
        cdmDatabaseSchema = CDM, #case sensitve
        vocabDatabaseSchema = Vocabulary,
        resultsDatabaseSchema = Results
      )
```

An alternative would be to convert to 
```

  daimons <- result$parsed$daimons
  daimons <- dplyr::bind_rows(daimons)
  daimons <- tidyr::spread(daimons, key = daimonType, value = tableQualifier)
  daimons <- dplyr::select(daimons, -sourceDaimonId)
  daimons[is.na(daimons)] <- ""
  daimons <- dplyr::group_by(daimons, sourceId)
  daimons <- dplyr::summarise_all(daimons, max)
  daimons <- dplyr::mutate_all(daimons, dplyr::na_if, "")
```

@schuemie because magrittr is such a useful package, i would like to propose that we add it to 'Depends' in the interest of code readability and ease of use. Infact, it would be great to add magrittr to the OHDSI R packages. in general..

gowthamrao: This is a good read

https://community.rstudio.com/t/magrittr-inside-a-package/2033

gowthamrao: Easiest solution seems to be to import the magrittr inflix operator for every function
```
#' @importFrom magrittr "%>%"
#'
```

There is another option with `usethis`
https://usethis.r-lib.org/reference/use_pipe.html

PR #30, opened 2020-01-10
rewrote function .checkBaseUrl

gowthamrao:
based on vignette for rex package here https://cran.r-project.org/web/packages/rex/vignettes/url_parsing.html

rex package https://cran.r-project.org/web/packages/rex/index.html

updated roxygen documentation and description file

discussion to support this change here
https://github.com/OHDSI/ROhdsiWebApi/pull/28#discussion_r364924527

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -15,17 +15,18 @@ License: Apache License 2.0
 Depends:
     R (>= 3.1.0),
 Imports:
-    RJSONIO,
-    httr (>= 1.3.1),
-    openxlsx (>= 4.0.17),
-    jsonlite,
     dplyr,
+    httr (>= 1.3.1),
+    jsonlite,
+    openxlsx (>= 4.0.17),
+    rex,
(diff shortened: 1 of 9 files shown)


schuemie:
A few comments:

1. It seems you have rewritten .checkBaseUrl function while maintaining its functionality. What I think you are going for is better readability of the code, but the price to pay is a dependency on the rex package. I think I prefer not having the dependency.

2. Have you made sure you can build the package? Does it pass R check? The reason I ask is that is still see `%>%`, which I don't think would work without having magrittr on the search path.

3. This might be a good opportunity to add a few unit tests that make sure this function works. You could test whether it detects some example malformed and well-formed URLs (see https://github.com/OHDSI/ROhdsiWebApi/blob/develop/tests/testthat/test-webapi.R)

alondhe:
@gowthamrao I would agree with @schuemie on the first comment especially. Are there cases in which the original function does not work correctly?

PR #32, opened 2020-02-25
changes to concept set

gowthamrao:
This PR creates a new function called 'resolveConceptSetId' that given a conceptSetId resolves the expression to return the included and mapped concepts as a data frame. This new function is useful to use included concepts/mapped concepts from a conceptSetId. Data frame is in camelCase.

Existing function createConceptSetWorkbook can now use this new function to create workbook.

Function has been tested/checked.


R/ConceptSet.R
```diff
@@ -332,17 +332,11 @@ resolveConceptSetId <-
...
-        header = FALSE,
-        sep = ",",
-        strip.white = TRUE,
-        blank.lines.skip = TRUE,
-        skipNul = TRUE
-      )
-      names(concepts) <- as.character(apply(concepts[1, ], 1, paste))
-      concepts <- concepts[-1, ]
+    .readCsv <- function(file) {
+      concepts <- readr::read_csv(file = file)
```
schuemie: I recommend using `col_types = readr::cols()` when calling read_csv to prevent the warning message that will confuse users.

Also, depending on the type of data in the CSV file, you may want to specifiy `guess_max = 1e7`, because else only the first 1,000 rows (I think) are used to guess the column types.

R/ConceptSet.R
```diff
@@ -332,17 +332,11 @@ resolveConceptSetId <-
...
+      names(concepts) <-
+        snakecase::to_lower_camel_case(names(concepts))
```
schuemie: I would prefer we standardize everywhere to using SqlRender to deal with case transformations.

R/Package.R
```diff
@@ -0,0 +1,24 @@
+# @file Package.R
```
schuemie: In all other OHDSI packages this file has the name of the package, in this case 'ROhdsiWebApi.R'.

schuemie:
Why is conceptSetExpression.xlsx in the root folder?

R/ConceptSet.R
gowthamrao: Thanks @schuemie this will require enhancement to SqlRender where it can convert an output with space such as 'this makes me happy' to 'thisMakesMeHappy'. SqlRender does not have this function currently, it can only convert snakeCase -> camelCase.

gowthamrao:
> Why is conceptSetExpression.xlsx in the root folder?

Mistake - needs to be deleted

R/ConceptSet.R
gowthamrao: That works! Thank you for the suggestion

R/ConceptSet.R
```diff
@@ -33,11 +33,14 @@
 #' @export
 getConceptSetName <- function(baseUrl, setId, formatName = FALSE) {
   .checkBaseUrl(baseUrl)
-
-  url <- gsub("@baseUrl", baseUrl, gsub("@setId", setId, "@baseUrl/conceptset/@setId"))
+  
+  url <-
+    gsub("@baseUrl",
```
alondhe: sprintf is a cleaner function for this: 
`sprintf("%s/conceptset/%s", baseUrl, setId)`

R/Package.R
```diff
@@ -0,0 +1,24 @@
+# @file ROhdsiWebApi.R
```
alondhe: What does this new file do?

gowthamrao: its modeled after https://github.com/OHDSI/SkeletonComparativeEffectStudy/blob/master/R/Package.R

it allows to provide meta information about the package. More importantly, it allows us to import packages/functions

alondhe: I'm not seeing this done in DatabaseConnector or SqlRender. @schuemie -- are we adding these Package.R files to packages now?

R/ConceptSet.R
gowthamrao: Maybe it is not part of this PR. It was already there?

I agree - we should use consistently. i like https://github.com/tidyverse/glue and find that to be better and easier to follow.

R/ConceptSet.R
schuemie: Well, in this case, why not create a simple custom function in this package, instead of adding the dependency?

R/Package.R
schuemie: @gowthamrao : I meant renaming the file, not just changing this line in the file.

@alondhe : Yes, all our R packages have such a file, for example [here](https://github.com/OHDSI/DatabaseConnector/blob/master/R/DatabaseConnector.R). But normally the file name is the same as the package name.

gowthamrao: Thanks. I made changes. Deleted package.R, modified ROhdsiWebApi.R

R/ConceptSet.R
gowthamrao: @schuemie i think this should do it. it worked in my tests
`string %>% 
  stringr::str_squish() %>% 
  stringr::str_to_lower() %>%
  stringr::str_to_title() %>%
  stringr::str_replace_all(pattern = " ", replacement = "")`

I committed the changes here (dependency on snakeCase package is now removed)
https://github.com/OHDSI/ROhdsiWebApi/pull/32/commits/5be55dfb90998141fd7e521ed60882b1a2c66f9a

R/ConceptSet.R
alondhe: My bad @gowthamrao -- didn't know this was already coded like this.

alondhe:
Hi @gowthamrao -- did you get my email suggesting an alternate way to build this functionality based on existing functions?

schuemie:
@alondhe , @gowthamrao : I badly needed some functionality for COVID-19, and went ahead and added it without consulting you. 

The up side is that it already provides the functionality @gowthamrao was looking for. See the example code [here](https://github.com/OHDSI/ROhdsiWebApi/blob/develop/extras/TestCode.R) for how you can now take a concept set either from within a cohort definition, or from WebAPI's list of concept sets, and output it either as (1) a cohort set expression table, (2) the list of included standard concepts, or (3) the list of mapped source concepts.

I noticed ROhdsiWebApi is growing in an odd way: for every output we require we build a function that gets JSON from WebAPI and outputs it to the required format. I propose a principle that I've attempted to implement in the new functions:

- There are functions for extracting (complex) R objects from WebAPI. These R objects no longer contain any JSON, but are full R object representations of the JSON, and can therefore be easily used in R, for example `cohortId <- estimation$cohortDefinitions[[1]]$id`
- These R objects can serve as input to functions that interact with WebAPI, which means in the back-end they're converted back to JSON (for example `conceptIds <- resolveConceptSet(conceptSet = conceptSet, baseUrl = baseUrl)`)
- We have separate functions that convert the R objects to the various required output formats, such as pretty tables, JSON, or anything else.

Let me know what you think

gowthamrao:
> * There are functions for extracting (complex) R objects from WebAPI. These R objects no longer contain any JSON, but are full R object representations of the JSON, and can therefore be easily used in R, for example `cohortId <- estimation$cohortDefinitions[[1]]$id`

Maybe we should ensure that these are convertible both ways - i.e. JSON -> R-object, R-object -> JSON. It would support use cases where we want to store/export a specification such as wanting to export (post) JSON back into Atlas (maybe another instance). I have found that in Atlas we are using nested JSONs, which makes this complex. So we will need to test this strategy.

> These R objects can serve as input to functions that interact with WebAPI, which means in the back-end they're converted back to JSON (for example conceptIds <- resolveConceptSet(conceptSet = conceptSet, baseUrl = baseUrl))

> We have separate functions that convert the R objects to the various required output formats, such as pretty tables, JSON, or anything else.

I have struggled with using R-objects as storage, but it makes it easier. Are you thinking of persisting the R-object in local drives like e.g. rds objects? Or just in-memory objects? 

are we thinking of using this same framework for data pulls too - e.g. extract (large) data into an R-object and then push to local dbms like sqlite (or ff?).  Or, is this framework limited to objects that are not data.

PR #34, opened 2020-03-16
Changed .checkBaseUrl to use the getWebApiVersion function

alondhe:
Rather than use a regex, we can simply check the WebAPI version via getWebApiVersion, and return TRUE if successful, and FALSE if that function errors.

### DIFF
--- a/R/WebApi.R
+++ b/R/WebApi.R
@@ -29,7 +29,7 @@
     stop("Base URL not valid, please verify it is like http://server.org:80/WebAPI")
   }
   
-  success
+  return(success)
 }
 
 .convertNulltoNA <- function(thisList) {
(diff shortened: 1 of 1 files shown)


R/WebApi.R
```diff
@@ -18,18 +18,18 @@
...
+    FALSE
   })
-  success <- any(as.logical(results))
 
   if (!success) {
-    stop("Base URL not valid, should be like http://server.org:80/WebAPI")
+    stop("Base URL not valid, please verify it is like http://server.org:80/WebAPI")
   }
+  
+  success
```
schuemie: Just nitpicking, but I prefer to be explicit about what is returned, so 

```r
return(success)
```

alondhe: No problem, just added it.

Issue #38, opened 2020-04-11
Function to delete/purge atlas artifacts


gowthamrao:
Given a list of atlas ids - we would like a function to delete/purge them. e.g. delete a list of cohort definition ids, incidence rate ids.

Justification: we routinely move/sync atlas definitions from one atlas instance to another on OHDSI network. Version control is a manual process. Many times we just want to purge old artifacts and replace them with new. Deleting/purging is a manual process e.g. imagine removing 20 cohort definitions one at a time.

Purging all atlas-ids for a project and replacing them would be a nice functionality.

Issue #37, opened 2020-03-31
Generic design pattern for ROhdsiWebApi


schuemie:
Continuing the discussion from [here](https://github.com/OHDSI/ROhdsiWebApi/pull/32#issuecomment-603764685)

So the proposal is that the main design pattern for this package is:
- Functions that go from WebApi to (comples) R objects that contain no JSON text. These objects can be inspected in R, for example to get the name of a concept or concept set.
- Functions that go from those R objects to the various required output formats, including tibbles, and where needed, JSON again.

This would reduce the growth of functions, which currently aim to go from WebApi directly to the desired output.

@gowthamrao: Yes, the idea would be the R objects are faithful representations of the JSON objects, so conversion back to JSON should be trivial. The new [`resolveConceptSet` function](https://github.com/OHDSI/ROhdsiWebApi/blob/develop/R/ConceptSet.R#L71) I created follows this pattern.

Could you explain what you mean when you say you "have struggled with using R-objects as storage"? Saving them as RDS objects works great, and if we want more portable objects, we can always saved them as JSON (at the time of saving). In fact, we use JSON to store R objects such as the CohortMethod analysis settings all the time. 

Could you describe uses cases where we need to pull large data objects from WebAPI? It helps to have specific scenarios in mind when discussing how we would approach them.

schuemie:
If we agree with this pattern, I suggest we delete:

- `getCohortDefinitionExpression` (because you could `getCohortDefinition` and then save as JSON. Maybe we do need a dedicated function for that operation)
- `getCohortDefinitionName ` (because you could `getCohortDefinition` and just get the name from the R object)
- `getConceptSetsAndConceptsFromCohort` (`getCohortDefinition`, then get the concept sets from the R object and call `resolveConceptSet`)

@alondhe , @gowthamrao , what do you think?

gowthamrao:
@schuemie I think we may need to add some functionality to get concept set expression from within cohort definitions.
- we need to get a list of concept set expressions with names - in a cohort
- we need to be able to post those concept set expressions and resolve them.

schuemie:
But we already have that functionality in R:
```r
cohortDefinition <- getCohortDefinition(cohortId = 14840, baseUrl = baseUrl)

conceptSet <- cohortDefinition$expression$ConceptSets[[1]]

conceptSet$name
# [1] "[OHDSI Cov19] Hepatitis C protease inhibitors"

[...]

PR #39, opened 2020-04-11
Adds ability to delete atlas ids using ROhdsiWebApi

gowthamrao:
Addresses https://github.com/OHDSI/ROhdsiWebApi/issues/38


schuemie:
Was the inclusion of `getCharacterizationDefinitionExpression` in this pull request intentional? It doesn't fit the issue description.

Also, the `getCharacterizationDefinitionExpression`  function doesn't follow the [proposed design pattern](https://github.com/OHDSI/ROhdsiWebApi/issues/37). We can decide to abandon that, but we'd need to have that discussion first.

gowthamrao:
Hi @schuemie  - yes, i should have explicitly stated inclusion of `getCharacterizationDefinitionExpression`

However, i think the output still follows the proposed convention. See return below. The reason is that `httr::content(json)' parses the object?

image

Issue #42, opened 2020-04-15
Post definitions into Atlas


gowthamrao:
We would like one or more functions to be able to post definitions such as cohort definitions into Atlas. 

get cohorts json expressoin
```
cohortExpression <- ROhdsiWebApi::getCohortDefinitionExpression(definitionId = definitionId,
                                                                baseUrl = baseUrl)
cohortJson <- cohortExpression$expression
```

Reviewing webApi documentation
http://webapidoc.ohdsi.org/index.html#398821407 
http://webapidoc.ohdsi.org/index.html#-847210453

PR #54, opened 2020-04-22
Adds function to post atlas definitions

gowthamrao:
Currently only supports cohort expression. TO DO: conceptSet, cohort characterization etc. 

tested
Thank you @anthonysena for the code

Addresses https://github.com/OHDSI/ROhdsiWebApi/issues/42


R/WebApi.R
```diff
@@ -237,4 +237,63 @@ getAtlasDefinitionsDetails <- function(baseUrl) {
     ) %>%
     dplyr::mutate(atlasCategory = SqlRender::camelCaseToTitleCase(atlasCategory))
   return(listOfAtlasIds)
+}
+
+
+
+#' Post a atlas study definition
```
schuemie, line 244: Several things here:
- atlas would need to be Atlas
- It is not necessarily Atlas but WebAPI
- Everything in this package is about WebAPI, so you really needn't mention Atlas or WebApi
- 'Study definition' is too broad. Is it an estimation study or a prediction study? 

(These comments obviously don't apply to just this line)

R/WebApi.R
```diff
@@ -237,4 +237,63 @@ getAtlasDefinitionsDetails <- function(baseUrl) {
...
+#' @examples
+#' \dontrun{
+#' validJsonExpression <- getCohortDefinition(baseUrl = baseUrl, cohortId = 15873)
+#' validJsonExpression <- RJSONIO::toJSON(cohortDefinition$expression)
+#' postAtlasDefinition(name = 'new atlas cohort name', 
+#'                      baseUrl = "http://server.org:80/WebAPI", 
+#'                      jsonExpression = validJsonExpression)
+#' }
+#' @export
+postAtlasDefinition <- function(baseUrl,
```
schuemie, line 270: Same problem here with Atlas, and now it has become just 'definition' (so no longer 'study'). This suggests I can use this function to post any definition?

schuemie:
I would think that if we stick to the pattern discussed earlier, the input to this function would be an R object, not a JSON string. The example code you provide already reads a bit awkward, in that need to translate to JSON ourselves from one ROhdsiWebApi function to the other. I recommend the new function translates to JSON internally.

R/WebApi.R
gowthamrao: Agree. I will rename the functions and documentation to not refer Atlas, but use webApi

> 'Study definition' is too broad. Is it an estimation study or a prediction study?
I agree there is ambiguity here. In this context i am referring to a definition of a cohort, concept set, incidence rate etc. A study would have many cohort definitions, concept set definition etc.

What general name should we use to refer to webapi (atlas) definitions? i.e. cohort definition, concept set definition etc?

R/WebApi.R
gowthamrao: You are right, i will address these in separate PR

schuemie:
I think we can close this pull request, right?

gowthamrao:
Closing this PR in favor of https://github.com/OHDSI/ROhdsiWebApi/pull/72

PR #55, opened 2020-04-22
Removing functions that do not follow design standards

gowthamrao:
https://github.com/OHDSI/ROhdsiWebApi/issues/37#issuecomment-617551140

addresses https://github.com/OHDSI/ROhdsiWebApi/issues/37

we maybe loosing some functionality


schuemie:
Would you mind deprecating `getCohortDefinitionExpression` and `getCohortDefinitionName` instead of deleting them? That way users will get warnings (nice) instead of errors (not nice) when they run their old code.

Issue #57, opened 2020-04-25
Support Patient Profile API call


gowthamrao:

gowthamrao:
Closed thru https://github.com/OHDSI/ROhdsiWebApi/pull/74

PR #58, opened 2020-04-25
Adds new function to get PersonProfile data

gowthamrao:
https://github.com/OHDSI/ROhdsiWebApi/issues/57


schuemie:
This pull request deviates from the [proposed design pattern](https://github.com/OHDSI/ROhdsiWebApi/issues/37) in that it not only returns the R object representation of the response from WebAPI, but also the unparsed JSON representation, which is called 'native'. What is the purpose of maintaining both? What use is there for an unparsed JSON expression of a patient's profile?

On another note, I would think the preferred way to get patient-level data is not through WebAPI but through a direct connection to the database. We have many R tools that work well against data in the CDM, including a [patient profile viewer](https://ohdsi.github.io/CohortDiagnostics/reference/launchCohortExplorer.html).

gowthamrao:
Maybe patient profile viewer is the way to go - and we dont need this function! I will look at patient profile viewer.

To address your comments:
> This pull request deviates from the proposed design pattern in that it not only returns the R object representation of the response from WebAPI, but also the unparsed JSON representation, which is called 'native'. 

The function `getPersonProfile` returns an R-object (not JSON) so i dont think it is a deviation. 

https://github.com/OHDSI/ROhdsiWebApi/blob/162b6cd5a45c67bcdcba8c0d5b8e12b9433b0988/R/PersonProfile.R#L87

> What is the purpose of maintaining both? What use is there for an unparsed JSON expression of a patient's profile?
The new private function `.getApiResponseParse` returns both parsed and json.  ways.https://github.com/OHDSI/ROhdsiWebApi/blob/162b6cd5a45c67bcdcba8c0d5b8e12b9433b0988/R/PersonProfile.R#L98
We can parse JSON many ways, and jsonlite seems to be very good for some types of JSONs with its `simplifyVector = TRUE, simplifyDataFrame = TRUE` , i.e. it takes care of converting NULL to NA, converting list of list into data-frame etc. . But converting this R-object back to JSON is not easy - I have struggled here (answer to your question [here](https://github.com/OHDSI/ROhdsiWebApi/issues/37#issue-590838112)) So, i had the option for retaining json in a private function.

R/PersonProfile.R
```diff
@@ -0,0 +1,115 @@
...
+#' @details
+#' Get a R object with person profile data. This function may be used for 
+#' visualizing a patients profile in tables or visualization.
+#'
+#' @param baseUrl        The base URL for the WebApi instance, for example:
+#'                       "http://server.org:80/WebAPI".
+#' @param sourceKey      The source key for a CDM instance in WebAPI, as defined in the
+#'                       Configuration page
+#' @param personId       The personId of the person whoose profile is being reviewed
+#' @param cohortId       <OPTIONAL> The cohortId of the cohort that the person belongs to
```
schuemie, line 30: What happens when I provide a cohortId? Do the cohort start and end date then become part of the patient profile? Please explain in the documentation.

PR #62, opened 2020-04-28
Incidence rate

gowthamrao:
https://github.com/OHDSI/ROhdsiWebApi/issues/59


R/IncidenceRate.R
```diff
@@ -0,0 +1,180 @@
...
+#' @return                   Get definitions for incident rate analysis
+#'
+#' @examples
+#' \dontrun{
+#' getIncidenceRateDefinitions <- function(baseUrl "http://server.org:80/WebAPI",
+#'                                          incidenceRateId = 296)
+#'                                          )
+#' }
+#' @export
+getIncidenceRateDefinitions <- function(baseUrl,
```
schuemie, line 36: The function name suggests plural, but the argument name (incidenceRateId) suggests singular. Why is this function returning multiple objects? (is it?) This would be different from say `getCohortDefinition`. 

I propose making it so this function returns a single incidence rate definition. If the user wants multiple, they can loop themselves. Unless there's a strong reason to stick to multiple? Are there multiple IR definitions with the same ID?

NAMESPACE
```diff
@@ -21,6 +21,9 @@ export(getConceptSetName)
 export(getConceptSetsAndConceptsFromCohort)
 export(getConcepts)
 export(getEstimation)
+export(getIncidenceRateDefinitions)
```
schuemie, line 24: I like having these 3 functions side-by-side. Maybe we should make this a pattern we apply consistently everywhere?

R/IncidenceRate.R
```diff
@@ -0,0 +1,180 @@
...
+
+#' Get incident rate generation information.
+#'
+#' @details
+#' Get incident rate generation information.
+#'
+#' @param baseUrl         The base URL for the WebApi instance, for example:
+#'                        "http://server.org:80/WebAPI".
+#' @param incidenceRateId The Atlas ID for incidence rate analysis.
+#' @return                An R-object
```
schuemie, line 61: Not very informative. I believe the R object in question is a tibble? Maybe say a little bit more about that tibble?

R/IncidenceRate.R
```diff
@@ -0,0 +1,180 @@
...
+#' @param outcomeCohortId    The cohort id of the outcome cohort in the incidence rate analysis.
+#' @return                   A list with incident rate results summary, stratified results and treemap data.
+#'
+#' @examples
+#' \dontrun{
+#' getIncidenceRateResults <- function(baseUrl "http://server.org:80/WebAPI",
+#'                                     incidenceRateId = 296,
+#'                                     targetCohortId = 432423, 
+#'                                     outcomeCohortId = 324,
+#'                                     sourceKey = "HCUP"
```
schuemie, line 115: Why a new line here?

gowthamrao: Where is the new line @schuemie

NAMESPACE
gowthamrao: Nice - the pattern being

- get definition (specification, expression as R object)
- get generation information (has it run?)
- get output (or results)

correct?

R/IncidenceRate.R
schuemie: Weird, it was there, but now it is gone

PR #65, opened 2020-04-30
Deprecating function getCohortDefinitionExpression and getCohortDefinitionName

gowthamrao:
https://github.com/OHDSI/ROhdsiWebApi/issues/37#issuecomment-617551140
https://github.com/OHDSI/ROhdsiWebApi/issues/37


schuemie:
Would you mind also mentioning in the function documentation that the function is deprecated? (i've been bad about this myself, but realize it would be nice for the user to know before calling the function).

Issue #16, opened 2019-09-01
getCohortInclusionRulesAndCounts - get both events and persons


gowthamrao:
Atlas has two tabs in cohort generation report
- By Events
- By Persons

It would be nice to have data from both of them returned by one function. This would help mirror the output in Atlas.

**Available in current version: from 'By Events' tab in Atlas**
$countSatisfying  -- 'N' (By Events)
$percentSatisfying -- %satisfied (By Events)
$percentExcluded -- %to-gain (By Events)
`
**Not Available in current version: from 'By Person' tab in Atlas**
'N' (By Persons)
'% remain' (By Persons)
'% difference' (By Persons)

There may be a way to calculate 'By Person' from current output, but i am not clear how to. It would be nice to get that pre-computed by the function.

anthonysena:
Another potential option here is to make the "mode" QueryString parameter of WebAPI an optional parameter to the `getCohortInclusionRulesAndCounts` function call. It can default to mode = 0 (By Event) and allow an override to mode = 1 (By Persons):

https://github.com/OHDSI/ROhdsiWebApi/blob/master/R/CohortDefinition.R#L516

gowthamrao:
Closed thru https://github.com/OHDSI/ROhdsiWebApi/pull/68

PR #68, opened 2020-04-30
Adds function to get results of cohort generation

gowthamrao:
Deprecating getCohortInclusionRulesAndCounts and replacing with getCohortGenerationReport that retains the output from deprecated functions and adds more functionality.

addresses https://github.com/OHDSI/ROhdsiWebApi/issues/16


R/CohortDefinition.R
```diff
@@ -634,3 +639,54 @@ deleteCohortDefinition <- function(cohortId, baseUrl, silent = FALSE, stopOnErro
...
+#' 
+#' @param baseUrl     The base URL for the WebApi instance, for example: "http://server.org:80/WebAPI".
+#' @param cohortId    The Atlas cohort definition id for the cohort
+#' @param sourceKey   The source key for a CDM instance in WebAPI, as defined in the Configuration page
+#' @param mode        Mode is used to differentiate between inclusion rules and count by events (mode = 0, default) 
+#'                    or persons (mode = 1). Default value = 0. 
+#' @return            A list of data frames containing cohort generation report
+#'
+#' @export
+getCohortGenerationReport <- function(baseUrl, cohortId, sourceKey, mode = 0) {
```
schuemie, line 656: This seems very similar to `getCohortGenerationStatuses`. Perhaps a more informative name? (I can't think of one, so maybe not)

gowthamrao: This function is returning the output of cohort generation process - so maybe `getOutputFromCohortGeneration` ?

but it wont follow the pattern in precedence `getCohortGenerationStatuses`

how about `getCohortGenerationOutput` - but it is not very far from `getCohortGenerationReport`. Output is a predecessor for Report - so maybe Output is a better choice vs report.

Issue #56, opened 2020-04-25
Use checkmate to provide informative error message


gowthamrao:
It would nice to use r-package [checkmate](https://cran.r-project.org/web/packages/checkmate/index.html) to manage errors/warning in ROhdsiWebApi. 

Similar to [SkeletonComparativeEffectStudy package](https://github.com/OHDSI/SkeletonComparativeEffectStudy/blob/develop/R/DataPulls.R)

PR #73, opened 2020-05-03
Added , add = errorMessage to the usage of checkerror

gowthamrao:
This allows to spool all checked errors and report them together, instead of one error at a time


gowthamrao:
Also addresses https://github.com/OHDSI/ROhdsiWebApi/issues/56

Issue #47, opened 2020-04-17
Add News.md


gowthamrao:
http://r-pkgs.had.co.nz/release.html

https://github.com/OHDSI/ROhdsiWebApi/issues/36#issuecomment-615282830

schuemie:
The NEWS.md file should be updated as part of the [release process]( https://ohdsi.github.io/MethodsLibrary/releaseProcess.html).

See for example [the ParallelLogger NEWS file](https://github.com/OHDSI/ParallelLogger/blob/master/NEWS.md).

gowthamrao:
Is news.md manually generated or using a script in say PackageMaintenance.R?

I also like how @anthonysena  manages these https://github.com/OHDSI/WebAPI/releases

schuemie:
The NEWS.md file must be manually created. We don't necessarily have a 1-on-1 relationship between the things in the NEWS file and issues. The NEWS file is to inform users of the package on what has changed.

PR #90, opened 2020-05-12
Create NEWS.md

gowthamrao:
https://github.com/OHDSI/ROhdsiWebApi/issues/47


NEWS.md
```diff
@@ -0,0 +1,25 @@
...
+	Ability to post definitions https://github.com/OHDSI/ROhdsiWebApi/issues/42
+	Use checkmate for informative error message https://github.com/OHDSI/ROhdsiWebApi/issues/56
+	Deprecated functions not following design pattern https://github.com/OHDSI/ROhdsiWebApi/issues/66
+	Estimation specification https://github.com/OHDSI/ROhdsiWebApi/issues/69
+	Retrieve a list of definition metadata https://github.com/OHDSI/ROhdsiWebApi/issues/40
+	Delete definitions https://github.com/OHDSI/ROhdsiWebApi/issues/38
+5. New ROxygen support https://github.com/OHDSI/ROhdsiWebApi/issues/44
+	
+	
+ROhdsiWebApi 0.0.3
```
schuemie, line 20: No point in adding old releases if we're not going to document the changes in them. Please remove.

NEWS.md
```diff
@@ -0,0 +1,25 @@
+ROhdsiWebApi 0.1.0
+=====================
+Note: this version has breaking changes because design changes and naming conventions.
+
+1. Generic design pattern: Based on discussions here https://github.com/OHDSI/ROhdsiWebApi/issues/37
```
schuemie, line 5: In general our OHDSI packages have two sections in NEWS (in line with CRAN expectations): Changes and Bug Fixes.

NEWS.md
```diff
@@ -0,0 +1,25 @@
...
+2. Consistent naming conventions: https://github.com/OHDSI/ROhdsiWebApi/issues/86 Use of WebApi instead of Atlas https://github.com/OHDSI/ROhdsiWebApi/issues/21 and case https://github.com/OHDSI/ROhdsiWebApi/issues/22
+3. Created Vignette
+4. New functions: 
+	Incidence rate https://github.com/OHDSI/ROhdsiWebApi/issues/59
+	Patient profile https://github.com/OHDSI/ROhdsiWebApi/issues/57
+	Ability to post definitions https://github.com/OHDSI/ROhdsiWebApi/issues/42
+	Use checkmate for informative error message https://github.com/OHDSI/ROhdsiWebApi/issues/56
+	Deprecated functions not following design pattern https://github.com/OHDSI/ROhdsiWebApi/issues/66
```
schuemie, line 13: Not a new function. Should be in a separate numbered item

NEWS.md
```diff
@@ -0,0 +1,25 @@
...
+	Estimation specification https://github.com/OHDSI/ROhdsiWebApi/issues/69
+	Retrieve a list of definition metadata https://github.com/OHDSI/ROhdsiWebApi/issues/40
+	Delete definitions https://github.com/OHDSI/ROhdsiWebApi/issues/38
+5. New ROxygen support https://github.com/OHDSI/ROhdsiWebApi/issues/44
```
schuemie, line 17: I don't think this is relevant for package users. Please delete

schuemie:
General comments: 

- Please use the two sections (Changes and Bug Fixes)

- My personal preference is not to have full URLs when linking to issues, but rather use hyperlinked words / numbers (can be added using markdown syntax)

- The news is very cryptic, and currently would require a user to click on every link you provide and then read a complex issue thread to understand. The news should be stand-alone. For example, instead of saying 'Generic design pattern', briefly explain what the design pattern is in a few sentences). Another example: instead of 'New functions: Incidence rates', provide the names of the new functions.

NEWS.md
gowthamrao: Should we consider v0.1.0 as the official first release? If yes, that would change the way we document News.Rd and we can start over.

i.e. there would be no 'changes' as this is the first release and we can start clean

gowthamrao: > Should we consider v0.1.0 as the official first release? If yes, that would change the way we document News.Rd and we can start over.
> 
> i.e. there would be no 'changes' as this is the first release and we can start clean

PR #95, opened 2020-05-15
Making naming more consistent. Refactoring

schuemie:
Tried to make function naming more consistent. Deleted some functions that didn't adhere to new design. Re-implemented some functions using newer functions. Making sure all functions return tibbles instead of data frames. Fixed some functions that weren't working.


R/CohortDefinition.R
```diff
@@ -685,20 +671,22 @@ getCohortGenerationOutput <- function(baseUrl, cohortId, sourceKey, mode = 0) {
...
     results$treemapData <- NULL
   } else {
-    treeMapResult <- list(name = c(), size = c())
-    treeMapResult <- .flattenTree(json$treemapData, treeMapResult)
-    json$treemapData <- dplyr::tibble(bits = treeMapResult$name, size = treeMapResult$size)
-    json$treemapData$size <- as.integer(json$treemapData$size)
-    json$treemapData$SatisfiedNumber = stringr::str_count(string = json$treemapData$bits, pattern = '1')
-    json$treemapData$SatisfiedRules = stringr::str_locate_all(string = json$treemaDpData$bits, pattern = '1') %>% 
-      paste()
+    results$treemapData <- jsonlite::fromJSON(json$treemapData, simplifyDataFrame = FALSE)
```
gowthamrao, line 680: wouldnt this return a tree of nested lists? 
.flattenTree was written to convert the nested tree to a dataframe

R/Estimation.R
```diff
@@ -43,30 +43,34 @@ getEstimation <- function(baseUrl, estimationId){
...
-  temp <- data$specification$estimationAnalysisSettings$analysisSpecification$targetComparatorOutcomes
-  
-  targetComparatorOutcomeIdDf <- data.frame(expand.grid(targetId = temp$targetId %>% unlist() %>% unique(),
-                                                        comparatorId = temp$comparatorId %>% unlist() %>% unique(),
-                                                        outcomeId = temp$outcomeIds %>% purrr::reduce(c) %>% unique())
-  )
-  data$specification$estimationAnalysisSettings$analysisSpecification$targetComparatorOutcomes <-
-    targetComparatorOutcomeIdDf
   
+  # If we don't change the R object, we can always convert it back to JSON
```
gowthamrao, line 58: Would the returned object be readily usable to user, or will more data wrangling needs to be done?

R/CohortDefinition.R
gowthamrao: Without .flattenTree - a nested list like this is returned - making it hard to use. 

image


.flattenTree helped convert the nested list to a dataframe

R/Estimation.R
gowthamrao: We may want to help with parsing the object. e.g. many times would want to expose the parameters chosen during analysis design choices - so that it may be used in R-markdown/reports/shiny/manuscript etc.

gowthamrao:
Hi @schuemie  I tested the package. 

To help with testing I made changes to `TestCode.R` file. During testing, I found a few issues (unrelated to this PR) - i have fixed some, added others to backlog.

Recommendation: I think results object should be returned as data-frame where possible. One of the changes you made (removing .flattenTree - returns a complex nested list of results. This is difficult to handle compared to data frame.) I think we should rework this part to incorporate a solution like .flattenTree that take such a complex nest and returns a more user friendly table format as described [here](https://github.com/OHDSI/ROhdsiWebApi/pull/95#discussion_r426259879)

schuemie:
Thanks @gowthamrao ! 

Like I said: the flattenTree code was just not working at all, throwing an error message. I tried to get it running, but couldn't figure it out.

I'm still on the fence on what to do with the objects pulled from WebAPI. In mind I make a distinction between 'assets' and 'status information':

1. An asset is something created in ATLAS, like a cohort definition, an estimation definition, etc., that has a specified structure. That structure is represented inside of WebAPI as a Java object. To transmit it, it is represented as a JSON object. In R we represent that same asset as an R object, still following the originally specified structure.

2. Status information is basically meta-information, like the status of a run, or the list of assets available. 

The assets are things we often want to preserve, and keep in the same structure, so we can convert them back from R to JSON back to the original Java objects. Therefore, we shouldn't apply all sorts of transformations to these objects, like flatten a tree structure. We could add a function that we can point at such an object to produce a pretty representation, but we'd have to keep the original without modification.

Because status information is transient by definition, there's no need to preserve as is, and it can be transformed to some pretty format (preferably a tibble) right away.

The reasons we don't want to maintain both the JSON and R representation of the same asset are:
1. Bad to have duplication.
2. How could I make changes to the object in R? Suppose I want to fetch a cohort definition, programmatically make some modificaction, and then upload it to WebAPI again? I don't want to manually edit the JSON string, but instead I'd want to modify the R object, and convert that to JSON.

schuemie:
I guess the results objects are a bit in-between. Maybe they should just be converted to nicely formatted R objects, since we would want to upload results back to WebAPI anyway.

gowthamrao:
Do you foresee use cases where we would want to post results object back
into WebApi? If not, maybe we could consider transforming results into
pretty format.

gowthamrao:
> An asset is something created in ATLAS, like a cohort definition, an estimation definition, etc., that has a specified structure. That structure is represented inside of WebAPI as a Java object. To transmit it, it is represented as a JSON object. In R we represent that same asset as an R object, still following the originally specified structure.

I like this description a lot @schuemie . I don't think `results` of an execution fit either of these categories. Example of results:  get counts of persons who met each rule in cohort definition. 
-  `results` are not `Assets`. The results are sourced by the WebApi from the underlying database that is storing the results. I don't think `results` are 'created in Atlas`.
- I don't think `results` are `Status` either - because it is not transient meta information. `results` are persisted in db.
>Status information is basically meta-information, like the status of a run, or the list of assets available.

I think `results' belong to their own category, and are the end product of the work. We could use `assets` and 'status' to make sense of the `results`. Since WebApi is just a transport medium from underlying db to end-user for `results` - we don't need to preserve the structure of the results. We also don't need to `post` results into WebApi. So - i don't think we 
>  want to preserve, and keep in the same structure, so we can convert them back from R to JSON back to the original Java objects. 

I propose, we deliver all `results` in table format (i.e. dataframe).

Issue #104, opened 2020-05-20
Ensure results objects are in table (data frame) format


gowthamrao:
See decision and discussion here https://github.com/OHDSI/ROhdsiWebApi/pull/95#issuecomment-631398515

PR #105, opened 2020-05-20
Results of cohort, are now returned as table

gowthamrao:
addresses .flattenTree https://github.com/OHDSI/ROhdsiWebApi/issues/104 and https://github.com/OHDSI/ROhdsiWebApi/pull/95#issuecomment-629994298

Also added a minor change to .checkBaseurl stop on error - message.

### DIFF
--- a/R/WebApi.R
+++ b/R/WebApi.R
@@ -25,11 +25,10 @@
   })
   
   if (!success) {
-    stop("Could not reach WebApi. Possibly the base URL is not valid or is not reachable? 
-         Please verify 
-          - is it in the form http://server.org:80/WebAPI,
-          - are you are connected to the network"
-         )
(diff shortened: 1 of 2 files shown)


R/WebApi.R
```diff
@@ -25,7 +25,11 @@
   })
   
   if (!success) {
-    stop("Could not reach WebApi. Possibly the base URL is not valid? (please verify it is like http://server.org:80/WebAPI)")
+    stop("Could not reach WebApi. Possibly the base URL is not valid or is not reachable?
```
schuemie, line 28: Lots of unnecessary spaces here. I recommend changing to:

```r
stop("Could not reach WebApi. Possibly the base URL is not valid or is not reachable?\n",
         "Please verify\n",
         "- is it in the form http://server.org:80/WebAPI,\n",
         "- are you are connected to the network")
```

Issue #109, opened 2020-05-26
Split getMetadataForAllSpecifications by categories and have getCategoryDefinitionsMetaData for each


gowthamrao:
https://github.com/OHDSI/ROhdsiWebApi/blob/6a117300b52633510c0f15f0b3f3de53c730fdf6/R/WebApi.R#L166

Related to https://github.com/OHDSI/ROhdsiWebApi/issues/108

function -   getMetaData | function - isValid
-- | --
getDataSourceDefinitionsMetaData | isDataSourceDefinitionsIdValid
getConceptSetDefinitionsMetaData | isConceptSetDefinitionsIdValid
getCohortDefinitionsMetaData | isCohortDefinitionsIdValid
getCharacterizationDefinitionsMetaData | isCharacterizationDefinitionsIdValid
getPathwayDefinitionsMetaData | isPathwayDefinitionsIdValid
getIncidenceRateDefinitionsMetaData | isIncidenceRateDefinitionsIdValid
getPersonDefinitionsMetaData | isPersonDefinitionsIdValid
getEstimationDefinitionsMetaData | isEstimationDefinitionsIdValid
getPredictionDefinitionsMetaData | isPredictionDefinitionsIdValid

Issue #108, opened 2020-05-26
Check WebApi for validity - we need functions to check if an id is valid/present in WebApi


gowthamrao:
Add the following new functions

function -   getMetaData | function - isValid
-- | --
getDataSourceDefinitionsMetaData | isDataSourceDefinitionsIdValid
getConceptSetDefinitionsMetaData | isConceptSetDefinitionsIdValid
getCohortDefinitionsMetaData | isCohortDefinitionsIdValid
getCharacterizationDefinitionsMetaData | isCharacterizationDefinitionsIdValid
getPathwayDefinitionsMetaData | isPathwayDefinitionsIdValid
getIncidenceRateDefinitionsMetaData | isIncidenceRateDefinitionsIdValid
getPersonDefinitionsMetaData | isPersonDefinitionsIdValid
getEstimationDefinitionsMetaData | isEstimationDefinitionsIdValid
getPredictionDefinitionsMetaData | isPredictionDefinitionsIdValid

@schuemie  what do you think?

PR #111, opened 2020-05-27
Autogenerated functions

gowthamrao:
Addresses https://github.com/OHDSI/ROhdsiWebApi/issues/109

https://github.com/OHDSI/ROhdsiWebApi/issues/108


schuemie:
The build is failing because the (auto-generated) examples are not marked as 'dontrun'. I'll fix that.

Issue #106, opened 2020-05-25
post concept sets to WebApi


gowthamrao:
https://forums.ohdsi.org/t/batch-uploading-concept-sets-into-atlas/10948/

gowthamrao:
Also with https://github.com/OHDSI/ROhdsiWebApi/pull/125

PR #116, opened 2020-05-28
Post concept set

gowthamrao:
Addresses https://github.com/OHDSI/ROhdsiWebApi/issues/106

With this PR we are able to post definitions of Cohorts and ConceptSets. 

- this code will check if a name assigned to the expression in the target WebApi - exists in the target WebApi (if yes, throws an error)
- posts the expression. If failed, it will return error message. If success, it will return details of the posted expression.

This framework is working for cohort and conceptset, but not incidence rate, pathways, cohort characterization, estimation or prediction. Tagging @chrisknoll @anthonysena  - as i may need guidance from you two

@schuemie could you please review this code


gowthamrao:
@schuemie i am having issues with R check and unit tests for now.

PR #118, opened 2020-05-30
Added getCohortDefinitionSqlFromExpression to autogenerate

gowthamrao:
Note:
- getCohortDefinitionSql will need to be deprecated.
- the default behavior of the new function is to generate SQL that includes cohort generation stats information where applicable.
- checks for WebApi version present in previous deprecated function has been removed, as it is only applicable to v 2.7.2
https://github.com/OHDSI/ROhdsiWebApi/blob/8ed2756f727753d3d031405b633c993369c5eb18/R/CohortDefinition.R#L305

See discussion here OHDSI#97

Closing OHDSI#100 in favor of OHDSI#97


schuemie:
Three things:
1. I think it is confusing that the argument here is the expression part of the definition, rather than the definition itself. Why not use the whole definition as input? (we could check under the hood if the input object has an `$expression` part. If not, we assume it is an expression itself.) 
2. The example is now useless, as it doesn't show you need to use the `$expression` part of the definition.
3. The name is unnecessarily long. I propose just `getCohortSql`

gowthamrao:
>  Why not use the whole definition as input?

My thoughts are that the whole expression is an artifact of how we have WebApi implemented i.e. needing metadata about name, created_by, description etc - which are not related to SQL generation process.

Also - as you described here https://github.com/OHDSI/ROhdsiWebApi/pull/120#issuecomment-636614793 and as we are investing in software like CHARYBDIS, we are going to be supporting more programmatic expression generation. In those cases, these meta data becomes not important or in few cases - conflicting/confusing.

I dont have a strong opinion either way - but this sounds a like another opportunity for a design decision.

R/AutoGeneratedDefinitions.R
```diff
@@ -407,6 +409,47 @@ invokeCohortDefinition <- function(cohortId, baseUrl, sourceKey) {
...
+#' \dontrun{
+#' getCohortSql(CohortDefinition = (getCohortDefinition(cohortId = 13242, baseUrl = baseUrl)),
+#'              baseUrl = "http://server.org:80/WebAPI")
+#' }
+#' @export
+getCohortSql <- function(cohortDefinition, baseUrl) {
+  .checkBaseUrl(baseUrl)
+
+  errorMessage <- checkmate::makeAssertCollection()
+  checkmate::assertList(x = cohortDefinitionExpression, min.len = 1, add = errorMessage)
```
schuemie, line 435: cohortDefinitionExpression is not defined

R/Deprecated.R
```diff
@@ -131,3 +131,57 @@ getCohortInclusionRulesAndCounts <- function(baseUrl, cohortId, sourceKey) {
...
+#'                        described in the details. By default this is TRUE.
+#'
+#' @return
+#' The templated SQL to generate the cohort
+#'
+#' @export
+getCohortDefinitionSql <- function(cohortId, baseUrl, generateStats = TRUE) {
+  .Deprecated(new = "getCohortDefinitionSql",
+              package = "ROhdsiWebApi",
+              msg = "This function has been deprecated. As an alternative please use getCohortDefinitionSqlFromExpression",
```
schuemie, line 159: Please update

schuemie:
Please update the one reference, and then feel free to merge

PR #119, opened 2020-05-31
Added fnction to check if a string name exits for a category

gowthamrao:


schuemie:
For functions that return a boolean, we typically use names like `isValid` or `existsConceptSetName`. I recommend we do the same here.

extras/templateWebApiFunctions.R
```diff
@@ -140,3 +140,35 @@ delete%categoryFirstUpper%Definition <- function(%category%Id, baseUrl){
...
+
+
+
+#' Check if %categoryFirstUpper% definition name exists.
+#'
+#' @details
+#' Check if a string name already exists in the WebApi as a %categoryFirstUpper% definition name.
+#'  
+#' @template BaseUrl
+#' @param %category%Name    A string name for the $categoryFirstUpper% to be checked.
```
schuemie, line 153: $ should be %?

schuemie:
The build fails as you can see [here](https://travis-ci.com/github/OHDSI/ROhdsiWebApi/builds/168999635#L978). 

I think this is because you have a `$` instead of a `%`, where indicated

Issue #35, opened 2020-03-20
Support WebAPI security


schuemie:
It is currently not possible to connect to a WebAPI that has security enabled. This may require some changes in WebAPI as well.

schuemie:
Seems to already be discussed [here](https://github.com/OHDSI/WebAPI/issues/1473#issuecomment-600176244)

alondhe:
There's a post on the WebAPI git that I think should be relevant here:
https://github.com/OHDSI/WebAPI/issues/1473#issuecomment-600103652

schuemie:
Jinx!

alondhe:
@schuemie beat me to it. Of course :)

gowthamrao:
thank you for working on this!

gowthamrao:
@ablack3 would you like to redo the PR based on the develop branch for v1.0.0 release?

ablack3:
@gowthamrao Yes!

ablack3:
@gowthamrao I'm actually not sure how to do this since I'm still fairly new to collaborating with git and github. I have commits on my develop branch that are not ohdsi branch. But there are commits on the OHDSI develop branch that I don't have. When I run `git pull upstream develop` I get a merge conflict. I aborted the merge. I'm also wondering if I do merge your changes into my develop repo and the create a  pull request would this mean that our develop branches will have different commit histories? How should I navigate this situation.

ablack3:
My WebAPI instance disappeared when I updated windows so it will take me some time to get everything set up again.

ablack3:
Talking with Konstantin about the the best way to add this. Since we want to support all types of authentication I think it is work thinking through how this should work. In every case we need to pass a header to the http requests. Sometimes that header is a token and other times it is a cookie. 
I think it would be if the user of this package did not need to manage the token or cookie. 
The user authenticates by running an `authenticate()` function which will allow the user to supply an auth method or guess the appropriate auth method if none is supplied. If authentication is successful an `authHeader` is created and saved in a temp file location that will be deleted when the R session ends. This header then automatically gets added to every `httr` call.
If the user gets a 401 error (unauthorized) on any call then a message prompting the user to call the `authenticate()` function will be displayed.

Or something like that. There is this idea of a function's or package's API (i.e. how the user of the package interacts with the functions) that I think is very important to think about.

ablack3:
I'm thinking of two possible approaches and open to direction or other ideas.

[...]

PR #127, opened 2020-06-08
Add option to use a bearerToken in getCdmSources.  #35

ablack3:
Adding a bearerToken argument to the getCdmSources function so that users can pass in a token and use this function. Github issue #35


gowthamrao:
Hi @ablack3, we are about to release v1.0 https://github.com/OHDSI/ROhdsiWebApi/milestone/2

Would it be ok if i review this PR after the release?

ablack3:
Yes of course!

gowthamrao:
@ablack3  do you want to try now? 

Also - i am assuming this will effect all functions?

PR #130, opened 2020-06-09
more tests

gowthamrao:


extras/PackageMaintenance.R
```diff
@@ -75,6 +75,7 @@ OhdsiRTools::formatRFile("R/AutoGeneratedDefinitions.R")
 
 
 ### test that code
+library(magrittr)
```
schuemie, line 78: dplyr subsumes magrittr, so if we depend on dplyr everywhere, we don't have to explicitly lists the magrittr dependency everywhere. 

In other words, I would just use `library(dplyr)` here.

extras/templateAutogeneratedTestFunctionScripts.R
```diff
@@ -20,6 +20,8 @@
 
 
 baseUrl <- Sys.getenv("ohdsiBaseUrl")
+library(magrittr)
```
schuemie, line 23: Same here

schuemie:
The build is currently failing because of the unlisted magrittr dependency. Also:
```
Non-standard files/directories found at top level:
  ‘NEWS.html’ ‘NEWS.Rmd’ ‘README.html’ ‘README.Rmd’
```

PR #136, opened 2020-06-11
Added vignette

gowthamrao:


DESCRIPTION
```diff
@@ -22,6 +22,7 @@ Imports:
     jsonlite,
     lifecycle,
     lubridate,
+    magrittr,
```
schuemie, line 25: Why do you need magrittr when dplyr is already imported?

PR #142, opened 2020-06-21
Add support for authentication using environment variable

ablack3:
This pull request explores an alternative approach to adding authentication using a global environment variable that holds the authentication header. This approach seems much simpler to implement than my previous pull request as it requires much less change to the code and should not break existing code. However there might be downsides to using global variables to store data and pass data between functions.

# Minimal example

``` r
library(ROhdsiWebApi)
baseUrl <- "http://localhost/WebAPI"
authenticateWebApi(baseUrl = baseUrl,
                   authMethod = "db",
                   webApiUsername = "user@domain.com",
                   webApiPassword = keyring::key_get("webApi"))


Sys.getenv("WEBAPI_AUTHHEADER")
#> [1] "Bearer eyJhbGcieiJIUzUxMiJ9.eyJzdW0iOiJhZG1gbkBkb21haW4uY2..."


cohort5 <- getCohortDefinition(5, baseUrl)
str(cohort5, max.level = 1)
#> List of 9
#>  $ id            : int 5
#>  $ name          : chr "[LEGEND HTN] Coronary Heart Disease"
#>  $ description   : NULL
#>  $ createdBy     : chr "user@domain.com"
#>  $ createdDate   : chr "2020-06-05 14:39"
#>  $ modifiedBy    : chr ""
#>  $ modifiedDate  : NULL
#>  $ expression    :List of 11
#>  $ expressionType: chr "SIMPLE_EXPRESSION"
```

<sup>Created on 2020-06-21 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


gowthamrao:
Hi @alondhe would this PR supersede this https://github.com/OHDSI/ROhdsiWebApi/pull/141 (if yes, could we close that one?)

ablack3:
> Hi @ablack3 would this PR supersede this #141 (if yes, could we close that one?)
It does not supersede #141. Both are draft proposals to be considered.  https://github.com/OHDSI/ROhdsiWebApi/issues/35#issuecomment-646829379
@gowthamrao What is your preference? Option 1 is probably better design but requires many more changes and breaks existing functions. Option 2 requires less changes but limits the user to one webApi session at a time and uses global variables to pass data between functions.

> Also - does this direction change the original intent to support multiple authentication methods?
No. This method would also support any authentication method.

ablack3:
This approach will not allow for interaction with multiple authorized WebAPI instances at the same time which is an important use case. I'm going to make another attempt modeling it after the gargle package.

Issue #143, opened 2020-07-02
postPathwayDefinition


hms1:
At the moment, `postDefinition` can only handle cohorts and concept sets. 

Modifications are needed to support `postPathwayDefinition`.

PR #144, opened 2020-07-02
Pathway upd

hms1:
This PR is to revolve [issue 143](https://github.com/OHDSI/ROhdsiWebApi/issues/143).

After these changes, `postPathwayDefinition` successfully posts a full pathways definition including constituent cohorts.

There may be a more elegant way to do this directly with `httr`, but I tried several ways and none seemed to work. 

The below method may also be applicable to other unsupported post functions that rely on importing cohorts nested within the study definition, for example PLE /characterization.

To summarize the changes:

1. The top level of the pathways json schema is different to cohorts and concept sets. It is handled by adding the below `R/PostDefinition.R` (along with some other small rearrangements):
```R
  if (category %in% c("pathway")){
    expression$name <- as.character(name)
    # convert R-object to JSON expression.  
    json <- RJSONIO::toJSON(expression)
  }  
```  

2. Posting a pathways json without having `eventCohorts` and `targetCohorts` already present in ATLAS (with the same IDs) previously failed. Particularly problematic was transferring pathways definitions between instances of ATLAS, where ID's are misaligned. In some cases the pathway appeared to post but uses irrelevant cohorts (same ID, different definition). `postPathwayDefinition` is changed  to post the cohorts before posting the pathway definition:  

```R
postPathwayDefinition <- function(name, pathwayDefinition, baseUrl, postCohorts = TRUE) {
  if ("expression" %in% names(pathwayDefinition)) {pathwayDefinition <- pathwayDefinition$expression}

  if(postCohorts){
    postModifyCohortDef <- function(cohortDef){
      output <- postCohortDefinition(cohortDef$name, cohortDef, baseUrl)
      cohortDef$name <- output$name
      cohortDef$id <- output$id
      return(cohortDef)
    }
    
    pathwayDefinition$targetCohorts <- purrr::map(pathwayDefinition$targetCohorts, postModifyCohortDef)
    pathwayDefinition$eventCohorts <- purrr::map(pathwayDefinition$eventCohorts, postModifyCohortDef)
  }
  
  result <- postDefinition(name = name,
                           baseUrl = baseUrl,
                           category = "pathway",
                           definition = pathwayDefinition)
  return(result)
}
```
3. `.getStandardCategories` is modified so that `getPathwayDefinition` pulls the full expression including cohort definitions. Without the change, `getPathwayDefinition`  just pulls the cohort IDs without the cohort definitions needed in the adapted version of `postPathwayDefinition`.

A couple of smaller changes not specific to pathways:

4. Added a few lines to `R/PostDefinition.R` to deal with names that have already been used in ATLAS (`name` is appended with (1) until the name has not been used):
   
```R
  categoryMetaData <- getDefinitionsMetadata(baseUrl = baseUrl, category = category)
  while(name %in% categoryMetaData$name){
    name <- paste0(name,"(1)")
  }
```
5. When `lubridate::guess_formats` returned more than one potential format, the `.convertToDateTime`  crashes. This PR modifies to pick the first potential format.


gowthamrao:
Hi @hms1 looks like Travis check failed. Could you please do a Rcheck locally 

See HADES guidelines here https://ohdsi.github.io/Hades/developerGuidelines.html
> Pull requests
> Before you do a pull request, you should always file an issue and make sure the package maintainer agrees that it’s a problem, and is happy with your basic proposal for fixing it. We don’t want you to spend a bunch of time on something that we don’t think is a good idea.
> 
> Additional requirements for pull requests:
> 
> Adhere to the Developer Guidelines posted here, as well as the OHDSI Code Style.
> 
> If possible, add unit tests for new functionality you add.
> 
> Restrict your pull request to solving the issue at hand. Do not try to ‘improve’ parts of the code that are not related to the issue. If you feel other parts of the code need better organization, create a separate issue for that.
> 
> Make sure you pass R check without errors and warnings before submitting.
> 
> Always target the develop branch, and make sure you are up-to-date with the develop branch

R/PostDefinition.R
```diff
@@ -62,12 +62,34 @@ postDefinition <- function(baseUrl, name, category, definition) {
...
-  json <- paste0("{\"name\":\"", as.character(name), "\",\"expressionType\": \"SIMPLE_EXPRESSION\", \"expression\":",
-                 jsonExpression,
-                 "}")
+  
+  categoryMetaData <- getDefinitionsMetadata(baseUrl = baseUrl, category = category)
+  while(name %in% categoryMetaData$name){
+    name <- paste0(name,"(1)")
+  }
+  
+  expression$createdDate <- NULL
```
gowthamrao, line 71: We need to design this to support all the atlas categories (i.e. characterization, estimation, cohort, concept set). Each of these have metadata like this createDate, modifiedDate etc.. but they are not consistent (esp characterization i think). So i used [this function](https://github.com/OHDSI/ROhdsiWebApi/blob/3e09d7350984c4ec9b4a93678b3492b5fcbf67ef/R/Private.R#L71) to change the returned object used [here ](https://github.com/OHDSI/ROhdsiWebApi/blob/3e09d7350984c4ec9b4a93678b3492b5fcbf67ef/R/GetDefinitionsMetadata.R#L71)

What you are doing is the opposite of what i did - trying to push the webapi, but the name should be change based on whether it is characterization or pathway or estimation etc. See 
[Characterization](http://webapidoc.ohdsi.org/index.html#1410315086)
image

[cohort-definition](http://webapidoc.ohdsi.org/index.html#398821407)
image

So, this PR needs to figure out how to change names based on what category is being posted. You could modify this [function](https://github.com/OHDSI/ROhdsiWebApi/blob/fe0a0d2978a891df4c8b23a639ec51c50f54c4f7/R/Private.R#L71) use the function like in [here](https://github.com/OHDSI/ROhdsiWebApi/blob/fe0a0d2978a891df4c8b23a639ec51c50f54c4f7/R/PostDefinition.R#L87)

R/PostDefinition.R
```diff
@@ -88,7 +110,13 @@ postDefinition <- function(baseUrl, name, category, definition) {
   response <- httr::content(response)
   structureCreated <- response
   response$expression <- NULL
-
+  
+  if(category %in% c("pathway")){
```
gowthamrao, line 114: I think we can support more categories using this framework

R/AutoGeneratedDefinitions.R
```diff
@@ -1565,7 +1566,26 @@ detectPathwaysByName <- function(pattern, negate = FALSE, baseUrl) {
 #'                       baseUrl = "http://server.org:80/WebAPI")
 #' }
 #' @export
-postPathwayDefinition <- function(name, pathwayDefinition, baseUrl) {
+postPathwayDefinition <- function(name, pathwayDefinition, baseUrl, postCohorts = TRUE) {
```
gowthamrao, line 1569: The AutoGeneratedDefinitions.R should be generated automatically in packageMaintenance 

see https://github.com/OHDSI/ROhdsiWebApi/blob/3e09d7350984c4ec9b4a93678b3492b5fcbf67ef/R/AutoGeneratedDefinitions.R#L18

and from [here](https://github.com/OHDSI/ROhdsiWebApi/blob/3e09d7350984c4ec9b4a93678b3492b5fcbf67ef/extras/PackageMaintenance.R#L52) 

I cant accept this change as written - because it did use existing framework and would break at next package update.

R/PostDefinition.R
```diff
@@ -62,12 +62,34 @@ postDefinition <- function(baseUrl, name, category, definition) {
...
   }
-  # convert R-object to JSON expression.
-  jsonExpression <- RJSONIO::toJSON(expression)
-  # create json body
-  json <- paste0("{\"name\":\"", as.character(name), "\",\"expressionType\": \"SIMPLE_EXPRESSION\", \"expression\":",
-                 jsonExpression,
-                 "}")
+  
+  categoryMetaData <- getDefinitionsMetadata(baseUrl = baseUrl, category = category)
+  while(name %in% categoryMetaData$name){
```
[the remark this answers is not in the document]
hms1: My concern is cases where a definition is impossible to import without manually editing names in the json definition (this could get messy, especially for definitions with embedded cohort definitions). Overwriting would solve this unless the existing definition is owed by another ATLAS username - in these cases overwriting may not be possible or appropriate.

Would you consider adding an argument to all postDefinition functions that specifies how to deal with name clashes?

Option 1 ('error') could be a meaningful error message. Option 2 ('overwrite') could attempt to overwrite. Option 3 ('rename') could rename with an information message giving details of the renaming.  It would make sense for Option 1 to be the default.

R/PostDefinition.R
hms1: Sorry - in retrospect the code I added here was lazy. I'll reverse engineer the existing function.

R/PostDefinition.R
hms1: Good to hear. I am happy to take a look at the remaining categories

R/AutoGeneratedDefinitions.R
hms1: Ah, I did wonder what the filename referred to.

So I guess the functionality here can be added to the main postDefinitions function early on (i.e. before the main `httr::POST`?

R/PostDefinition.R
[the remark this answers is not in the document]
gowthamrao: Yes, adding such an argument would be useful for sure. Agree with option 1 being default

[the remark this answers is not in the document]
hms1: A```duplicateNames``` argument  has been added to ```post%categoryFirstUpper%Definition``` in ```R/templateWebApiFunctions.R``` and hence all post functions in ```R/AutoGeneratedDefinitions.R```. 

The argument can take values ```c("error", "overwrite", "rename")``` and is passed into the ```postDefinition``` function, where it determines the behavior of ```.checkModifyDefinitionName```.

R/PostDefinition.R
hms1: When the correct pathways expression is used, forcing ```NULL``` values is not necessary - removed.

R/AutoGeneratedDefinitions.R
hms1: I have moved this functionality from ```R/AutoGeneratedDefinitions.R``` (which is now generated directly using ```PackageMaintanence.R``` into ```postDefinition```:

https://github.com/hms1/ROhdsiWebApi/blob/8a6a7ab6f61fad6c3b978cf4af7ba67d7ee70935/R/PostDefinition.R#L77-L89

hms1:
Not sure why this is failing travis checks.. something to do with the ```UsingROhdsiWebApi``` vignette it does not like:

```
* creating vignettes ...Error in find_vignette_product(name, by = "weave", dir = docdir, engine = engine) : 
  Failed to locate ‘weave’ output file ‘UsingROhdsiWebApi.pdf’ or ‘UsingROhdsiWebApi.html’ for vignette with name ‘UsingROhdsiWebApi’ and engine ‘utils::Sweave’. The following files exist in working directory ‘/tmp/Rtmp9ucSK7/Rbuild31254f5cbd85/ROhdsiWebApi/vignettes’: ‘UsingROhdsiWebApi.R’ (NA bytes), ‘UsingROhdsiWebApi.Rmd’ (NA bytes)
Execution halted
```

hms1:
Going to close this PR and submit a new one since a lot has changed since the original PR. Scope will also change to include additional categories

Issue #121, opened 2020-06-01
Use caching for some operations


schuemie:
Some of the same calls to WebAPI will be done repeatedly, for example to see if the baseURL is correct, or to retrieve the list of valid IDs. We can avoid this, thus not only making the functions faster but also avoid unnecessary calls to the WebAPI, by using caching.

One way to implement this would be to create a global environment like [here](https://github.com/OHDSI/Andromeda/blob/develop/R/Andromeda.R#L28), and use that to store a cache. 

For example, we only need to test whether a baseURL is valid once. We can store the list of valid baseURLs in the environmental variable, and check the list at the next call.

gowthamrao:
Since WebApi supports multiple concurrent users, the cache may get obsolete. How do we make sure it does not become obsolete..

gowthamrao:
After discussing with @schuemie  - we should only use it for baseUrl (which should not change on concurrent changes).

ablack3:
We may also want to cache something related to user authentication. At the very least we should cache the bearer token. We might also want to cache a refresh token that would allow ROhdsiWebApi to automatically get a new bearer token when the current one expires. Does WebApi grant refresh tokens?

gowthamrao:
@ablack3 @schuemie  is this still an issue that we need to work on?

schuemie:
I would definitely cache whether the baseUrl is valid. That would save a full round-trip for each call, and if it becomes outdated you have another problem ;-) 

If I recall correctly, most calls also retrieve the full list of valid IDs in order to see if the one the user requested is among them. That is a very expensive operations (especially as the number of valid IDs grows.). I don't actually see the point in that behavior, because if you didn't do that then the user would just get an 'invalid ID' error from the WebAPI itself a few milliseconds later, which would be just as informative. But if you want to keep that behavior, I recommend caching the list of valid IDs at least for a minute or so, so if I loop over IDs (for example to get multiple cohort definitions) I don't have to download the full list of valid IDs every time.

PR #148, opened 2020-07-09
Add authentication/authorization for db and Active Directory auth mentods

ablack3:
This pull request adds support for WebAPI security using the db and Active Directory auth methods (#35 )

It provides a framework for adding support using other methods as the details are worked out. It also provides a mechanism to cache results for multiple WebAPI instances through the use of a package global environment.  (#121) Anything we want to cache can be stored in a named list in this environment. The name of the list is the baseUrl which allows for multiple WebAPI instances. There will be one list per baseUrl/WebAPI instance. This implies that baseUrls must uniquely identify a WebAPI instance (no trailing slashes).

These commits implement the following changes

1) Add global package environment that will cache bearer tokens
2) Add wrappers to httr request funtions to centralizing all http requests and http error handling
3) Add authorization function that is called by users to generate and store a WebApi auth token
4) fix small error in PostDefinition.R

Since this PR deals with security it should be reviewed by someone with a security background. A few things to consider from the security perspective:
- This change will allow ROhdsiWebApi to send a username and password to WebApi and receive a token. The token is cached for the duration of the R session in memory (not on disk). 
- The username and password are not saved anywhere which I think makes this safer than `DatabaseConnector::createConnectionDetails()` which in typical use does temporarily save passwords.
- Once the token expires the user needs to re-authenticate and get a new token
- Caching the token during the R session seems like a minimum requirement for the desired functionality. 

While this PR looks like a lot of changes it is not actually very large. Most of the files that have changed are auto generated documentation  that I can remove from the PR. Two files have been added with the authorization functionality and two files have been added that are used to test the new functions. httr methods have been replaced with internal wrappers throughout the package which is why so many files have changed. 

I have trouble getting R CMD Check to pass for reasons unrelated to this PR. I think the reason that R CMD Check does not pass is because the vignette cannot be built. Building the vignette relies on an external server and I was getting http errors when trying to build the vignette.

> SUMMARY: processing the following file failed:
>   ‘UsingROhdsiWebApi.Rmd’

Specifically the build error is caused by 

> We can now post this R-object into WebApi as follows:
> ```{r eval=TRUE, echo = TRUE}
> returnFromPostRequest <- ROhdsiWebApi::postConceptSetDefinition(baseUrl = baseUrl, 
>                                                                 conceptSetDefinition = rExpression, 
>                                                                 name = conceptSetName)
> ```
> 
> Status code = 500 [Internal server error]
> Error in postDefinition(name = name, baseUrl = baseUrl, category = "conceptSet",  :


ablack3:
I've been using this branch and it seems to be working well. Still having trouble with the vignette though.

ablack3:
Hi @gowthamrao Any chance I can get this PR into version 1.1? Is the Travis CI check the only issue?

chrisknoll:
Tagging @schuemie , as he may have familiarity with this error, since he's very good at building vignettes, and may be familiar with this error.

gowthamrao:
@ablack3  @chrisknoll @schuemie I think i know the vignette issue, 

@chrisknoll based on your prior experience with implementing webApi security - could you help us by reviewing this PR

Team - i think this PR requires a separate branch - i will create a branch and merge this PR to that

R/CohortDefinition.R
```diff
@@ -281,3 +281,8 @@ getCohortSql <- function(cohortDefinition, baseUrl, generateStats = TRUE) {
     stop()
   }
 }
+
+#' @export
+print.cohort <- function(x, ...) {
```
gowthamrao, line 286: why make it exportable

R/httrWrappers.R
```diff
@@ -0,0 +1,78 @@
+# Internal wrappers for httr request functions 
+# 
+# The primary reason for this wrapper is to add the appropriate authorization header to requests.
+# This file also centralizes http error handling.
+GET <- function(url, config = list(), ..., handle = NULL) {
```
gowthamrao, line 5: maybe we use .GET, .POST, .PUT instead of just GET, POST, PUT

to prevent potential name conflict with HTTR::GET, HTTR:POST

gowthamrao:
@chrisknoll @schuemie would you be able to review?

@ablack3  have you tested this on webapi's with and without security enabled?
What auth methods have you tested?

gowthamrao:
@ablack3  after @chrisknoll and/or @schuemie  review the code changes. i will have to ask @anthonysena  to please help do some testing. If its ok with you, i would like to , after their review merge your contribution to new branch.

After testing, and making sure there is no other auth method that needs to be added using your framework

i propose we merge it to develop and plan to release it as ROhdsiWebApi v1.2

chrisknoll:
The large drop in code coverage (55%) is a problem, and there there should be tests that validate the new code.  I don't have extensive experience in how this test could be implemented, but I imagine some sort of mock that can simulate a WebAPI auth endpoint could be implemented to ensure the proper interactions with WebAPI are performed given the authentication method.

gowthamrao:
Agree with @chrisknoll  , @ablack3 are you able to add tests that improve the code coverage. For a package, higher the better - but for HADES we strive for a minimum 55% to be part of ohdsi Hades. This package is part of HADES.

Also, after approval of this PR but before release to master - I think a vignette related to authorization would be needed. You already have good content in the issue post #35

ablack3:
@gowthamrao, Thanks for the feedback and the fix to the vignette.

> @ablack3 have you tested this on webapi's with and without security enabled?
> What auth methods have you tested?

I only have access to instances of webapi with security enabled so I have not tested it on instances without security. I think it will work fine but it should be tested. I have tested internal database security (default method for OHDSIonAWS) and Active Directory. Both methods are essentially the same from the point of view of ROhdsiWebApi.

There is a function to manually set the bearer token for a specific URL. `setAuthHeader()` which allows the user to use any method as long as they can get a bearer token by whatever means they like. I have an example of this using Google's IAP here: https://github.com/OHDSI/ROhdsiWebApi/issues/35#issuecomment-674531883

These changes support multiple WebAPI instances and I have used it to copy cohorts from one Atlas instance to another. Google's oauth has been a real challange and I think the reason has to do with the redirect url. However if/when we figure that out it could be added without much trouble. I think the major change to the package is the use of the httr wrappers. 

> @ablack3 are you able to add tests that improve the code coverage?
> I imagine some sort of mock that can simulate a WebAPI auth endpoint could be implemented to ensure the proper interactions with WebAPI are performed given the authentication method.
I'll have to think about how to do this and give it a shot. I think ideally we would test against a real WebAPI instance. 
Do tests need to be able to be run from any computer?

R/httrWrappers.R
ablack3: Yes I like this idea. I'll make the change.

R/CohortDefinition.R
ablack3: You're right. I removed the export tag.

ablack3: Actually I removed this change altogether since it is not relevant to this PR. Adding generic methods for ROhdsiWebApi objects is discussed in https://github.com/OHDSI/ROhdsiWebApi/pull/150

gowthamrao:
@ablack3 thanks for making the changes. Lets wait for review by @chrisknoll , @schuemie  or @anthonysena . 

After they review, i would like to merge it to authenticationSecurity branch. Would you be ok if i 'squash and merge' or do you have preference to keep every commit history change?

chrisknoll:
It should be a general practice to squash the commits in the PRs before merge to keep the history clean.  The individual commits from the PR will still linger in the repo for some time if there is any interest in it, but there shouldn't be a need to see all the revisions and edits when the history should be clearly defining what changed for a specific feature.

gowthamrao:
Great! thanks @chrisknoll i will plan to squash and merge

ablack3:
Sounds like a plan! I’m working out an idea to create a simulated webapi endpoint using the plumber package and spawn it in a separate process. Then tests would use that plumber api endpoint.

ablack3:
Converted this PR back to draft status so I can finish adding tests.

ablack3:
@gowthamrao, I think this is ready for another look. I believe that my new code is mostly covered by new tests using a simulated webapi endpoint created with plumber.

schuemie:
Sorry, I'm not going to review 132 changed files. If you guys are ok with it, then go ahead.

schuemie:
@gowthamrao: To move this pull request forward, might I suggest:

1. Testing this on https://atlas.ohdsi.org/, which seems to have security enabled
2. Making sure this PR doesn't break anything for people that don't use security.

ablack3:
> 1. Testing this on https://atlas.ohdsi.org/, which seems to have security enabled
Unfortunately https://atlas.ohdsi.org/ uses Google authentication only (as far as I know) and this PR does not implement security for Google/oauth/openID.

schuemie:
Ok, well, that leaves step 2 then ;-)

azimov:
@schuemie @chrisknoll I think this PR can move forward.

1. Testing this on https://atlas.ohdsi.org/, which seems to have security enabled

So I just tested this on https://atlas.ohdsi.org by pulling the token out of the WebApi requests and it works.

    token <- "Bearer somLongString copied from your browser dev console"
    setAuthHeader("https://atlas.ohdsi.org/WebApi", token)
    cohort <- getCohortDefinition(406, "https://atlas.ohdsi.org/WebAPI") # Works
    
 So that's a crude solution that would work for users that currently can't use this atlas instance...

The auth for could be added with the googleAuthR package, so that's an option if we want to support google oAuth in an R session. I couldn't get this working for me locally but I can spend more time and this could be added as a feature at a later date.

2. Making sure this PR doesn't break anything for people that don't use security.

Running this branch on our insecured instance appears to work fine. The httrWrappers.R functionality is pretty strightfoward and handles requests for baseUrls with no authentication (as well as multiple atlas instances) cleanly.

chrisknoll:
Ok, will take a look. Btw: we can add a feature to atlas 2.9 that will let you copy your auth token to clipboard so that it's easier for users to get the token out of atlas and into their R session.

azimov:
that would be a useful feature (and it only needs to be written in javascript)

schuemie:
@azimov: are you ok with merging this PR?

azimov:
> @azimov: are you ok with merging this PR?

I think the code is fine but would like the recent changes from develop for github actions to be merged in first, as well as it pointing at develop rather than the authenticationSecurity feature branch. I don't have permissions on the branches to do this though. @ablack3 can you update the PR?

schuemie:
Package owner @gowthamrao can do those things. @gowthamrao : we recommend accepting this pull request., but first accepting the one switching to GitHub Actions.

gowthamrao:
Excellent! Thank you @schuemie  @azimov  for reviewing the PR! I am merging the PR

gowthamrao:
Thank you @ablack3  @azimov @chrisknoll @schuemie 

@ablack3 this is an excellent contribution!

PR #165, opened 2020-09-24
Files seems to be misplaced

gowthamrao:
These files seem to be misplaced and caused cohort diagnostics to fail


schuemie:
These files are used by [insertCohortDefinitionSetInPackage()](https://github.com/OHDSI/ROhdsiWebApi/blob/master/R/CohortDefinition.R#L111) function, which inserts them in the study package (E.g. [here](https://github.com/OHDSI/ROhdsiWebApi/blob/master/R/CohortDefinition.R#L211)). If you move them, the function won't work. 

We could argue over whether the SQL files need to be in the inst/sql folder, but I view that folder for being for SQL that SqlRender needs to translate (using the `loadRenderTranslate()` function). In this case the SQL just needs to be copied.

gowthamrao:
Thanks @schuemie this code in Cohort Diagnostics breaks without the changes and were in unit test
https://github.com/OHDSI/CohortDiagnostics/blob/818ea0d4559d5b382b3ee67cfcb29d5f3b44aea5/R/CohortConstruction.R#L771

So we will have to either change in cohort diagnostics or change here

PR #174, opened 2021-03-05
Add vignette on WebAPI authorization from R

ablack3:
This PR adds a new vignette that explains how to authorize WebAPI from R so that ROhdsiWebApi functions can be used with security-enabled WebAPI instances. Issue #173


azimov (approved):
Looks good and reads pretty clearly to me.

PR #177, opened 2021-03-17
added windows auth

azimov:
Draft PR to add windows auth (which is different to active directory) to the authentication layer

Draft, needs a unit test.

Only tested manually on mac osx using string username and password the code below should work :
```
authorizeWebApi(baseUrl, "windows", webApiUsername = "Myusername", webApiPassword = askpass::askpass("Web api password")) # OS X

authorizeWebApi(baseUrl, "windows", webApiUsername = ":", webApiPassword = ":")) # Windows
```


R/Auth.R
```diff
@@ -36,7 +36,8 @@ authorizeWebApi <- function(baseUrl, authMethod, webApiUsername = NULL, webApiPa
   # run appropriate auth. Each auth method must return a header to be added to WebAPI calls.
   authHeader <- switch(authMethod,
                        db = .authDb(baseUrl, webApiUsername, webApiPassword),
-                       ad = .authAd(baseUrl, webApiUsername, webApiPassword))
+                       ad = .authAd(baseUrl, webApiUsername, webApiPassword),
+                       windows = .authWindows(baseUrl, webApiUsername, webApiPassword))
```
anthonysena, line 40: I know this is still in draft but just thought I'd ask as I looked through these changes. Should this be: `.authWindows(baseUrl, ":",":")` instead?

azimov: That doesn't work on a OS X which is where I wrote the code. I'm not sure it will work on windows either as I'm yet to test it there.

anthonysena:
I'll give this a test on Windows

R/Auth.R
anthonysena: Ah good to know. I'll give it a try on Windows as-is to see how it goes.

gowthamrao:
@azimov @anthonysena lemme know if your testing is success and we are ready to merge

anthonysena:
@azimov - `authorizeWebApi(baseUrl, "windows", webApiUsername = ":", webApiPassword = ":")` worked well for me on Windows. Passing in ":" as the username/password is required by `httr` but it certainly feels like something I'd forget to do when aiming to use Windows auth. I'm wondering if that there should be some type of handling such that when authMethod== 'windows' and `webApiUsername` and `webApiPassword` are NULL, we set webApiUsername/webApiPassword to ":" to save some typing. Then it will allow users that want to explicitly pass their credentials a way to do this. 

@gowthamrao - I'd hold on this until @azimov gives the green light. The good news is that it is functional but would welcome suggestions on the Windows auth point above.

azimov:
> @azimov - `authorizeWebApi(baseUrl, "windows", webApiUsername = ":", webApiPassword = ":")` worked well for me on Windows. Passing in ":" as the username/password is required by `httr` but it certainly feels like something I'd forget to do when aiming to use Windows auth. I'm wondering if that there should be some type of handling such that when authMethod== 'windows' and `webApiUsername` and `webApiPassword` are NULL, we set webApiUsername/webApiPassword to ":" to save some typing. Then it will allow users that want to explicitly pass their credentials a way to do this.
> 

@sena good idea I this in a recent commit. Windows users with nt login enabled should just be able to do:

```
authorizeWebApi(baseUrl, "windows")
```
But this will error out on other platforms. I added a description for this in the vignette.

If future webapi versions could expose an endpoint that gave login options we could present the user with auth options when a request fails. E.g:

```
401 error: ...

Your web API has security enabled. Please select an option from the following
1. Windows domain login
2. Active Directory login
3. Database login
4. Google/OAuth login (currently not supported, see Vignette("...") for more details

Contact your ATLAS/WebAPI administrator for more details... Blah blah blah
```


I have also added test cases to support real web api instances that can be set with environment variables. These can be supported in future github actions but, for now, this allows testing on other platforms with live instances. The tests are skipped where these environment variables are not found so we can get good coverage now before needing any test webapi instances to be configured.

@gowthamrao the file `test-live-webapi.R` is the basis for converting your test code to unit tests. However, I didn't want to pollute peoples live atlas versions with junk cohort definitions. Modifying `test-webapi.R` might be a better solution as this has all the boilerplate that is needed. However, it says it was auto-generated so i didn't want to alter it.

tests/testthat.R
```diff
@@ -1,2 +0,0 @@
-library(testthat)
-test_check("ROhdsiWebApi")
```
anthonysena, line 2: Isn't this file required per https://r-pkgs.org/tests.html#test-workflow or perhaps there is a newer approach?

R/Auth.R
```diff
@@ -20,10 +21,21 @@
...
   errorMessage <- checkmate::makeAssertCollection()
   checkmate::assertCharacter(baseUrl, len = 1, min.chars = 1, add = errorMessage)
-  checkmate::assertChoice(authMethod, choices = c("db", "ad"), add = errorMessage)
+  checkmate::assertChoice(authMethod, choices = c("db", "ad", "windows"), add = errorMessage)
+
+  # With windows type we can try NT user authentication
+  if (authMethod == "windows" &
+    is.null(webApiUsername) &
+    is.null(webApiPassword) &
+    .Platform$OS.type == "windows") {
```
anthonysena, line 34: Nice check!

anthonysena (commented):
These changes look good from my side @azimov. Since this is still in draft, I left a small question on testthat.R. If you feel like this is ready for merge, please let @gowthamrao know and then we can get this merged. Thanks!

tests/testthat.R
azimov: I took this pattern from packages that use DatabaseConnector on different platforms. Ideally we would have a WebApi test instance that supports Db, active directory windows etc login platforms so the files `testAdAuth.R testWindowAuth.R, testDbAuth.R` represent moving towards that goal... but I can revert this if you think an alternative pattern is better.

anthonysena: Fair enough - if what you've done allows for unit testing, I'm good with this change.

gowthamrao:
@anthonysena or @schuemie could you please review the PR and approve for merge

tests/testthat.R
gowthamrao: @anthonysena do you approve the changes. Are we ok to merge?

anthonysena (approved):

tests/testthat.R
anthonysena: Yes, just left the approval. Thanks!

Issue #152, opened 2020-08-13
RJSONIO::toJSON loss of precision


anthonysena:
An example of this: when serializing a cohort definition, a concept set included in the definition had this concept and concept_id:

siltuximab
concept_id == 44818461

The default settings for RJSONIO::toJSON() change this formatting of this concept in JSON to use scientific notiation thus resulting in the value:

4.481846e+07

When re-hydrating this definition that contains scientific notation, the concept_id is then incorrectly specified as `44818460` instead which is a different drug. 

The solution (which was found in the code base) is to use the `digits` formatting extension:

RJSONIO::toJSON(data, **digits = 23**)

gowthamrao:
@anthonysena do you think it would be useful to show a working example? If true, this is a major high priority issue and will require 'deep cleansing'

anthonysena:
Here is one:

``` r
foo <- list(expression = list(ConceptSets = list(list(expression = list(items = list(list(concept = list(CONCEPT_ID = 44818461))))))))
foo$expression$ConceptSets[[1]]$expression$items[[1]]$concept$CONCEPT_ID
#> [1] 44818461
RJSONIO::toJSON(foo, pretty = TRUE)
#> [1] "{\n\t\"expression\" : {\n\t\t\"ConceptSets\" : [\n\t\t\t{\n\t\t\t\t\"expression\" : {\n\t\t\t\t\t\"items\" : [\n\t\t\t\t\t\t{\n\t\t\t\t\t\t\t\"concept\" : {\n\t\t\t\t\t\t\t\t\"CONCEPT_ID\" : 4.481846e+07\n\t\t\t\t\t\t\t}\n\t\t\t\t\t\t}\n\t\t\t\t\t]\n\t\t\t\t}\n\t\t\t}\n\t\t]\n\t}\n}"
RJSONIO::toJSON(foo, digits = 23, pretty = TRUE)
[...]

PR #179, opened 2021-03-19
reverted to old use of jsonlite which formats JSON objects from data frames very differently.

azimov:
This was creating an issue with `postCohortDefinition`

E.g.
```
cohort_definition <-
  getCohortDefinition(cohortId = 15997, baseUrl = baseUrl)
postCohort <-
  postCohortDefinition(
    name = paste0("this is part of a test and may be deleted ", tempfile() %>% basename()),
    cohortDefinition = cohort_definition,
    baseUrl = baseUrl
  )
```

Would throw a 400 error.

This PR reverts to use of jsonlite from RJSONIO which was changed in another commit.

The issue is with how R data frames are encoded. For example:

```
> data <- data.frame(a=1,b=2)
> RJSONIO::toJSON(data)
[1] "{\n \"a\": [        1 ],\n\"b\": [        2 ] \n}"

> jsonlite::toJSON(data)
[{"a":1,"b":2}] 
```

I've reverted to using Json lite because a) this is the format WebApi expects b) its still used elsewhere in the codebase and c) their transformation of data frames to a json string is probably more efficient and better tested than anything we should bother trying to write quickly.

### DIFF
--- a/R/PostDefinition.R
+++ b/R/PostDefinition.R
@@ -101,7 +101,7 @@ postDefinition <- function(baseUrl, name, category, definition) {
                     .data$isExcluded,
                     .data$includeMapped,
                     .data$includeDescendants)
-    expression <- RJSONIO::toJSON(x = items, digits = 23, pretty = TRUE)
+    expression <- jsonlite::toJSON(x = items)
     responsePut <- .putJson(url = paste0(baseUrl,
                                          "/",
                                          argument$categoryUrl,


gowthamrao:
We ran into precision issue - with using jsonlite

```
library(magrittr)
smallNumber <- 3243234
bigNumber <- 342432432432432432
```

`jsonlite::toJSON(x = smallNumber) %>% jsonlite::fromJSON() %>% scales::comma()`
"3,243,234"

`jsonlite::toJSON(x = bigNumber) %>% jsonlite::fromJSON() %>% scales::comma()`
"342,432,432,432,432,000"

`RJSONIO::toJSON(x = smallNumber) %>% RJSONIO::fromJSON() %>% scales::comma()`
"3,243,234"

`RJSONIO::toJSON(x = bigNumber) %>% RJSONIO::fromJSON() %>% scales::comma()`
"342,432,400,000,000,000"


**precision set with digits = 23 (no such option in jsonlite0)**
`RJSONIO::toJSON(x = smallNumber, digits = 23) %>% RJSONIO::fromJSON(digits = 23) %>% scales::comma()`
"3,243,234"

`RJSONIO::toJSON(x = bigNumber, digits = 23) %>% RJSONIO::fromJSON(digits = 23) %>% scales::comma()`
"342,432,432,432,432,448"


This level of precision has become important as we are now regularly using bigint in the range.

gowthamrao:
See https://github.com/OHDSI/ROhdsiWebApi/issues/152

gowthamrao:
Also btw - i just tested the code example above.

i am able to post with v1.2 (in develop) - current develop version

azimov:
> i am able to post with v1.2 (in develop) - current develop version

Hmm, I definitely still have this problem after pulling the latest version of the develop branch.

In this post the user is putting concept identifiers limited by the vocabulary, will these ever exceed the digit limit? I thought this problem is for users that want to create identifiers that are like `CONCEPT_ID * 10000`. Is this possible in the cohort definitions given the current concept ids?

I agree that we should probably move to RJSONIO in the medium term but I don't think that this use case is justified here given that it breaks working code for users and we currently don't have a working build that will give us good test cases.

gowthamrao:
@azimov this is not a jsonlite issue, but a R precision issue. The big number example here does not fit R's numeric range.

azimov:
> @azimov this is not a jsonlite issue, but a R precision issue. The big number example here does not fit R's numeric range.

I understand, RSJONIO puts the data frame in a different format, not expected by web api. The latest commit by transposing the data by hand before posting.

R/PostDefinition.R
```diff
@@ -101,7 +101,12 @@ postDefinition <- function(baseUrl, name, category, definition) {
                     .data$isExcluded,
                     .data$includeMapped,
                     .data$includeDescendants)
-    expression <- RJSONIO::toJSON(x = items, digits = 23, pretty = TRUE)
+
+    itemsTranspose <- apply(items, 1, function(item) {
+      list(item)
+    })
+
+    expression <- RJSONIO::toJSON(x = itemsTranspose, digits = 23, pretty = TRUE)
```
anthonysena, line 109: I'd revise this to use the .toJSON() function that is private to this package as detailed [here](https://github.com/OHDSI/ROhdsiWebApi/blob/master/R/Private.R#L177).  This way the magic "23" is not floating around the code base.

anthonysena (changes requested):

anthonysena (approved):

gowthamrao:
Great! Thank you @anthonysena  for reviewing. I will merge now

PR #190, opened 2021-05-31
Switching to tempEmulationSchema

schuemie:
Use of oracleTempSchema will throw deprecation warning.

The newer versions of SqlRender and DatabaseConnector use tempEmulationSchema instead of oracleTempSchema, and this schema can be set via a global option.


azimov (approved):
Looks good, the check that fails appears to be an issue with the plumbR mock api

PR #227, opened 2021-09-02
Remove non-ascii characters from WebAPI response

anthonysena:
Per discussion on OHDSI/OhdsiRTools#34, I'm proposing a change to remove non-ASCII characters from WebAPI's response by default. 

Please note this is a first-draft with some questions & outstanding items:

1. Should we enforce this default for all responses? The issue referenced above is mainly when writing the response to the file system for a study package. 
2. This PR is missing unit test(s) to verify the requirement of removing non-ASCII characters


anthonysena:
I'm going to close this out and handle this operation in CohortGenerator.

PR #231, opened 2021-10-27
Update definitions

### LINKED ISSUE #146: Modify definition
Currently we can get and post definitions, but we cannot modify definitions. We would like ability to modify definitions e.g. change name, change an parameter etc.. This maybe achieved by replacing an old definition with a new one.

azimov:
Resolves #146

Changes made:
* Added updateDefinition that follows the webapi/atlas workflow for updating definitions
* Added template scripts
* Updated auto generated function calls

If there are no major issues with this change I would like to include it in release 1.2.1


azimov:
I removed the auto-generated code for updates other than cohorts and concept sets.

A notable difference in the way this code works is that it does not use mocks for the unit tests as this does not appear to work (httptest doesn't like that the stored json changes in the middle of the test), so these query the test server directly.

R/GetPostDeleteDefinition.R
```diff
@@ -253,6 +253,84 @@ postDefinition <- function(baseUrl, name, category, definition) {
...
+
+  errorMessage <- checkmate::makeAssertCollection()
+  checkmate::assertCharacter(definition$name, add = errorMessage)
+  checkmate::assertCharacter(category, add = errorMessage)
+  checkmate::assertNames(x = category, subset.of = arguments$categoryStandard)
+  checkmate::assertInteger(definition$id)
+  checkmate::reportAssertions(errorMessage)
+
+  # Check entity exists - this will throw a meaningful error if it doesn't
+  tryCatch(
```
gowthamrao, line 286: small change request for formatting - i think our OHDSI HADES convention is 

`tryCatch({`

not

```
tryCatch(
{
```

R/GetPostDeleteDefinition.R
```diff
@@ -253,6 +253,84 @@ postDefinition <- function(baseUrl, name, category, definition) {
...
+    stop(paste(definition$name, "Is not a valid cohort name."))
+  }
+
+  # Check definition is ok
+  if (category == "cohort") {
+    jsonExpression <- .toJSON(definition$expression)
+    checkUrl <- paste(baseUrl, argument$categoryUrl, "check", sep = "/")
+  
+    tryCatch(
+      {
```
gowthamrao, line 314: image

gowthamrao (changes requested):
Just formatting changes please

gowthamrao (approved):

PR #236, opened 2021-12-09
Implement exportCohortSet

chrisknoll:
Added method to accept a set of cohortIDs and a WebAPI url that will fetch the cohort definitions and OHDSI-sql for the specified cohort definitions.

Tests are included.


NEWS.md
```diff
@@ -1,4 +1,3 @@
-
 ROhdsiWebApi 1.2.1
 ==================
```
azimov, line 3: This sould have a line like "implemented export of cohort definition sets for study packages"

R/exportCohortDefinitionSet.R
```diff
@@ -0,0 +1,57 @@
...
+      cohortName = character(),
+      sql = character(),
+      json = character(),
+      logicDescription = character(),
+      generateStats = logical()
+    )
+    
+    for (i in (1:length(cohortIds))) {
+      cohortId <- cohortIds[i]
+      ParallelLogger::logInfo(paste("Fetching cohortId:", cohortId))
```
azimov, line 33: As stated above, I would prefer a call to `message` here to remove the dependency.

DESCRIPTION
```diff
@@ -28,6 +28,7 @@ Imports:
     RJSONIO,
     scales,
     SqlRender,
+    ParallelLogger,
```
azimov, line 31: While I don't think this will be an extra dependency for most users, this is only used in a single function call so I would suggest removing it

azimov (approved):

azimov (approved):

PR #238, opened 2021-12-09
Fixes #226.

### LINKED ISSUE #226: getConceptSetDefinition returns empty expression if the concepts are not in the priority vocabulary_key
In our Atlas configuration we have CDM databases with different vocabularyies.

If i use `getConceptSetDefinition()` to get the concept_set, the field expression apears empty. 
This also happens if I use directly the webAPI : 
"http://localhost:8080/WebAPI/conceptset/[id]/expression"

```json
{"items":[]}
```

This does not happen if the concepts in the concept_set are in the priority vocabulary 
OR
If i do 
"http://localhost:8080/WebAPI/conceptset/[id]/expression/[sourceKey]"

But  `getConceptSetDefinition()`  does not allow me to se the sourceKey

Could we include the parmeter `vocabularySourceKey` in `getConceptSetDefinition()` in the same way that it appears in `resolveConceptSet()`

chrisknoll:


chrisknoll:
Due to how the other getDef functions were being templated, I wasn't able to have the existing getConceptSetDefinition accept a vocab source key as a param.  Instead, I had to introduce a new function name to handle that usecase.

tests/testthat/test-getConceptSetDefinitionBySourceKey.R
```diff
@@ -0,0 +1,7 @@
+with_mock_dir("mocks/getConceptSetDefinitionBySourceKey", {
+  
+  test_that("getConceptSetDefinitionBySourceKey works", {
```
azimov, line 3: is it possible to have a test with `vocabularySourceKey` set to some value?

chrisknoll: Yep!

PR #241, opened 2022-01-18
Added addCohortPermission.

chrisknoll:
Removed validation check from cohort saving because you should be allowed to save a cohort definition even with errors.


R/GetPostDeleteDefinition.R
```diff
@@ -302,18 +302,21 @@ updateDefinition <- function(definition, baseUrl, category) {
...
+  # # Check definition is ok 
   
-    tryCatch({
-      response <- .postJson(checkUrl, jsonExpression)
-      content <- httr::content(response)
-    }, error = function (error) {
-        stop(paste("Error with", category, "definition:", error))
-    })
-  }
+  ## why??  can't you save a definition with errors?
```
azimov, line 307: I'm happy to allow cohorts to be posted if there are errors (as long as ATLAS allows this - there should be consistency between the javascript and R clients), but the user should be presented with the warnings/notes when this happens.

R/permissions.R
```diff
@@ -0,0 +1,50 @@
...
+#' @param userId      The userId to grant permission
+#' @param permission  The permission type to assign.  Can be READ or WRITE. Defaults to "WRITE".    
+#'
+#' @examples
+#' \dontrun{
+#' getCohortSql(CohortDefinition = (getCohortDefinition(cohortId = 13242, baseUrl = baseUrl)),
+#'              baseUrl = "http://server.org:80/WebAPI")
+#' }
+#' @export
+addCohortPermission <- function(baseUrl, cohortId, userId, permission="WRITE") {
```
azimov, line 35: Does this function work with older versions of atlas (i.e. is there an informative error if the permission endpoints don't exist?)

azimov (changes requested):

R/permissions.R
chrisknoll: Not exactly: the problem is that there was a API breaking change in atlas 2.10 related to the call to `check`, so killing two birds here:  even with errors, you should be allowed to save cohorts, and b) removing a call to `check` that fails tests.  I can remove this code block if it's distracting, but I figured I'd just call it out.

ablack3:
> Note, I am not able to perform security related tests due to that we don't have a security-enabled dev site to test with.

I think there is (or was) a security enabled webapi instance to test against. These are the environment variables I was using. I think Lee set this up. I'm not sure if these variables are available on the github actions runners.

WEBAPI_TEST_WEBAPI_URL
WEBAPI_TEST_SECURE_WEBAPI_URL
WEBAPI_TEST_ADMIN_USER_NAME
WEBAPI_TEST_ADMIN_USER_PASSWORD

R/GetPostDeleteDefinition.R
ablack3: I'm confused. This tryCatch is handling http errors which seem different than errors with the cohort. For example, error code of 500 "internal server error" should not mean "the cohort definition has a problem but the POST operation succeeded" right?

ablack3:
@chrisknoll  What is the preffered way to check if security/auth is enabled in WebAPI?

I'm also thinking we need a way to get user IDs using ROhdsiWebAPI.

chrisknoll:
Yes, if you do a HTTP GET to `WebAPI/info` you get this back (this is from my local env:)

```json
{
  "version": "2.12.0",
  "buildInfo": {
    "artifactVersion": "WebAPI 2.12.0-SNAPSHOT",
    "build": "NA",
    "timestamp": "Fri Jul 29 18:11:47 UTC 2022",
    "branch": "issue-2061",
    "commitId": "a4dbb85",
    "atlasRepositoryInfo": {
      "milestoneId": 42,
      "releaseTag": "*"
    },
    "webapiRepositoryInfo": {
      "milestoneId": 43,
      "releaseTag": "*"
    }
  },
  "configuration": {
    "security": {
      "samlActivated": false,
      "enabled": false,
      "samlEnabled": false
    },
    "vocabulary": {
      "solrEnabled": false
    },
    "plugins": {
      "atlasgisEnabled": false
    },
    "person": {
      "viewDatesPermitted": false
    },
    "heracles": {
      "smallCellCount": "5"
    }
  }
}
```

Under configuration -> security -> enabled, you can determine if security is enabled. Note: the /WebAPI/info endpoint should not require an authentication header.

ablack3:
Here is a test for the new function. @leeevans I think we need an account that can set permissions on the security enabled webapi instance. 

``` r
# devtools::install_github("OHDSI/ROhdsiWebApi", ref = "permissions-cohortDef")

library(ROhdsiWebApi)

baseUrl <- Sys.getenv("WEBAPI_TEST_SECURE_WEBAPI_URL") 

authorizeWebApi(baseUrl = baseUrl,
                authMethod = "db",
                webApiUsername = Sys.getenv("WEBAPI_TEST_ADMIN_USER_NAME"),
                webApiPassword = Sys.getenv("WEBAPI_TEST_ADMIN_USER_PASSWORD"))

df <- getUserPermission(baseUrl) 

names(df)
#> [1] "id"          "login"       "name"        "permissions"

# We need an account that can be used to test this function
setCohortPermission(baseUrl, cohortId = 3, userId = 1000, permission = "WRITE")
#> Error in `.request()` at ROhdsiWebApi/R/httrWrappers.R:10:2:
#> ! http error 403: Forbidden request.
#>    You do not have permission to perform this action.
```

<sup>Created on 2022-07-29 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

ablack3:
I've made some edits to this PR. Let me know what you all think. I like the get/set pattern for modifying permissions. I don't have a great understanding of how webapi handles permissions though.  Currently we have 

`getUserPermissions` which hits the users endpoint http://webapidoc.ohdsi.org/index.html#-741996078
`setCohortPermissions` which hits the permissions endpoint http://webapidoc.ohdsi.org/index.html#439635542

I'm not quite sure what the pattern should look like but I do think we need a way to get User IDs and get the permissions set by the function being added in this PR.

leeevans:
@ablack3
The test admin user has the Atlas admin role, it has 'cohort creator' role and I've added it to a new role called 'set permissions' with all the permissions that contain the word 'permission'.
When I try
setCohortPermission(baseUrl, cohortId = 3, userId = 1000, permission = "WRITE")
It still fails with a permissions error.

I can update the permissions for the test admin user, if you can let me know what specific set of permissions it needs.

Additional info:
I logged into the secure atlas as the test admin user and I was able to successfully manually update and save cohort id 3. I undid the change I made in atlas and saved it again successfully. 

I also manually made a copy of cohort id 3 in Atlas, which is cohort id 559 (owned by the test admin user) and the following command executed without an error:
setCohortPermission(baseUrl, cohortId = 559, userId = 1000, permission = "WRITE")

(I temporarily installed a copy of this branch in my local RStudio for testing)

ablack3:
@gowthamrao has asked for help with this PR and I can put a little time toward it to try and help out. I'd like to get a clear idea of the scope of this PR. Usually we discuss the scope of a change in an issue and I'm not able to find an issue associated with the PR. From what I see in the code the goal here is to expose some WebAPI permission management through ROhdsiWebApi.

In particular it seems that @chrisknoll would like to be able to set cohort read/write permissions. I think we also need a way to get user permissions so we can see the result of the change from ROhdsiWebApi.

What would be helpful for me is some clear documentation about how WebAPI handles permissions. It seems there are roles which can be granted READ or WRITE permissions on various entities (cohort, pathway, etc). A user ID can be used as a role ID. 

My general question with respect to this PR is about scope. Is our goal here for ROhdsiWebAPI to have functions to get/set Cohort level permissions for any user ID? Would it make more sense to expose more general permission management through ROhdsiWebApi that mirrors what is available through WebAPI or are we only interested in user level cohort permissions?

R/GetPostDeleteDefinition.R
gowthamrao: @chrisknoll  :) cant understand your concern. 

This is not checking if json has errors. It is only checking if it posted it.  All this try catch is doing is - replacing default error message with something different

gowthamrao (changes requested):

R/GetPostDeleteDefinition.R
gowthamrao: > There is centralized http error code handling in the .request function which is being overridden here.

This function was written before the support for security. When we decided to support security in ROhdsiWebApi - we did not remove these old error handling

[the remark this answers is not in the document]
chrisknoll: Why can't you save a cohort definition if it has validation errors in it?  Preventing saving a cohort definition with validation errors makes little sense to me....

gowthamrao:
Closing in favor of https://github.com/OHDSI/ROhdsiWebApi/pull/254

Issue #246, opened 2022-03-10
Add getUsersFromRole function


alondhe:
Add a function to obtain the users mapped to a WebAPI role.

PR #247, opened 2022-03-10
Add getUsersFromRole function

alondhe:
#246 enhancement - function to get users from a WebAPI role, for security enabled Atlas instances


gowthamrao:
Hi @alondhe i like the PR, but i am finding it hard to do a review. Mainly because it seems to be bundled with formatting and many changes. There are changes in extras folder. Plus - there are potential merge conflicts etc.

Apologies - but would you be willing to do a clean PR which just the targeted changes?

PR #254, opened 2022-08-29
Permission related functions

gowthamrao:
Because this https://github.com/OHDSI/ROhdsiWebApi/pull/241 was hanging around for sometime, i am going to cherry pick and add to develop branch

Also going to copy from https://github.com/OHDSI/ROhdsiWebApi/pull/247 @alondhe


gowthamrao:
I really wanted the functionality to give users permission to cohorts (large number), so i visited PR related to that. But couldnt get it to work. So i thought maybe cherry pick some of the other functions and put it in develop.

Then revisit the set permission function

gowthamrao:
I added the setCohortPermission as experimental. 
It is "working", but i am confused if it is assigning properly

R/permissions.R
```diff
@@ -0,0 +1,216 @@
...
+    
+    if (!permission %in% c("READ", "WRITE")) {
+      stop(paste0(
+        "Invalid Permission Type:",
+        permission,
+        ".  Valid types are READ or WRITE."
+      ))
+    }
+    
+    url <-
```
gowthamrao, line 175: http://webapidoc.ohdsi.org/index.html#417537417

POST http://www.example.com/permission/access/entityType/entityId/role/roleId/

is userId the same as roleId -- i dont think so.

e.g. roleId =1 is public. I can assign WRITE permissions to roleId 1!

gowthamrao: This morning i was able to assign WRITE permission (cant do READ yet) to another user using this funcitn

gowthamrao:
After thinking this thru, talking with @anthonysena and experimenting  - it turns out setCohortPermission is actually setting permission to a roleId , not a userId

So - we need to change the functions parameters.

Doing so - makes this work.

gowthamrao:
Documentation is here http://webapidoc.ohdsi.org/index.html#417537417

chrisknoll:
Per this discussion: https://github.com/OHDSI/ROhdsiWebApi/pull/241#discussion_r796115849, the original PR removed the validation check from saving a cohort definition  (I'll ask here again: why does it make sense to prevent saving a cohort definition that has validation errors??)

That was half the point of the PR I raised in the first place, but this PR supersedes it?

